PR #8, opened 2020-11-24
revisit queue usage + debug failure handling

richardliaw:


xgboost_ray/main.py
```diff
@@ -293,6 +293,7 @@ def _train(params: Dict,
            checkpoint_prefix: Optional[str] = None,
            checkpoint_path: str = "/tmp",
            checkpoint_frequency: int = 5,
+           enable_queue: bool = False,
```
richardliaw, line 296: untested... may remove later

xgboost_ray/main.py
```diff
@@ -342,27 +345,35 @@ def _train(params: Dict,
...
-            for i, (actor, queue) in enumerate(actors):
+            if queue:
                 while not queue.empty():
                     item = queue.get()
                     if isinstance(item, Callable):
                         item()
                     else:
                         callback_returns[i].append(item)
-            ready, not_ready = ray.wait(fut, num_returns=len(fut), timeout=0)
+            ready, not_ready = ray.wait(not_ready, timeout=0)
```
richardliaw, line 362: This is a key difference - this makes it so that we do not wait for all workers to finish - rather, every worker that finishes, we call `ray.get` on its result. This will allow error messages to propagate properly, even before the rest of the workers are done.

xgboost_ray/main.py
```diff
@@ -342,27 +345,35 @@ def _train(params: Dict,
...
+            time.sleep(1)
+            ray.get(ready)
         # Once everything is ready
         ray.get(fut)
     except RayActorError:
-        for (actor, _) in actors:
+        for actor in actors:
             ray.kill(actor)
+        if queue:
+            ray.kill(queue)
```
richardliaw, line 372: this is also important to kill, otherwise we end up leaking a bunch of queues (which may not necessarily garbage collect properly).

xgboost_ray/main.py
```diff
@@ -133,8 +133,8 @@ def __init__(self,
                  checkpoint_path: str = "/tmp",
                  checkpoint_frequency: int = 5):
         self.queue = queue
-
-        init_session(rank, queue)
+        if queue:
+            init_session(rank, queue)
```
krfricke, line 137: Shouldn't we always want to initialize the session with the rank?

xgboost_ray/main.py
```diff
@@ -342,27 +345,35 @@ def _train(params: Dict,
...
                 while not queue.empty():
                     item = queue.get()
                     if isinstance(item, Callable):
                         item()
                     else:
                         callback_returns[i].append(item)
-            ready, not_ready = ray.wait(fut, num_returns=len(fut), timeout=0)
+            ready, not_ready = ray.wait(not_ready, timeout=0)
+            logger.debug("[RayXGBoost] Waiting for results...")
+            time.sleep(0.5)
```
krfricke, line 364: Is this `sleep` necessary?

xgboost_ray/main.py
krfricke: Good spot thanks.
Should we catch broader here?

xgboost_ray/main.py
krfricke: That makes sense, as other workers are waiting for the killed worker to finish before they finish themselves. That explains why it stalled

xgboost_ray/main.py
krfricke: I'll update the code below

krfricke (approved):
Thanks, that's great. I'll merge it and update the minor things I left in the comments

PR #20, opened 2020-12-01
Better checkpointing/fault tolerance + tests

krfricke:
Better checkpoint handling. Instead of writing local checkpoints, this PR sends checkpoints from one worker via the queue to the driver. This is effectively checkpointing via the object store, thus no longer requiring local temp files.

It is not straightforward to load a checkpoint into Rabit, so this way seems to be the most versatile way to make sure each worker has access to the latest checkpoint.

Todos (possibly in follow-up PRs): 
- Multinode release test
- Save last checkpoint on driver on fail if maximum number of restarts is exceeded

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -383,6 +383,7 @@ def _train(params: Dict,
            evals=(),
            ray_params: RayParams,
            _checkpoint: _Checkpoint,
+           _additional_results: Dict,
            **kwargs) -> Tuple[xgb.Booster, Dict, Dict]:
     _assert_ray_support()
 
@@ -458,7 +459,11 @@ def _train(params: Dict,
         for actor in actors
     ]
 
-    callback_returns = [list() for _ in range(len(actors))]
+    callback_returns = _additional_results.get("callback_returns")
+    if callback_returns is None:
+        callback_returns = [list() for _ in range(len(actors))]
+        _additional_results["callback_returns"] = callback_returns
+
     try:
         not_ready = fut
         while not_ready:
@@ -486,10 +491,9 @@ def _train(params: Dict,
     res: Dict[str, Any] = ray.get(fut[0])
     bst = res["bst"]
     evals_result = res["evals_result"]
-    additional_results = {}
 
     if callback_returns:
-        additional_results["callback_returns"] = callback_returns
+        _additional_results["callback_returns"] = callback_returns
 
     all_res = ray.get(fut)
     total_n = sum(res["train_n"] or 0 for res in all_res)
@@ -499,7 +503,7 @@ def _train(params: Dict,
 
     _shutdown(remote_workers=actors, queue=queue, force=False)
 
-    return bst, evals_result, additional_results
+    return bst, evals_result, _additional_results
 
 
 def train(params: Dict,
@@ -563,6 +567,7 @@ def train(params: Dict,
 
     tries = 0
     checkpoint = _Checkpoint()
+    current_results = {}
     while tries <= max_actor_restarts:
         try:
             bst, train_evals_result, train_additional_results = _train(
@@ -572,6 +577,7 @@ def train(params: Dict,
                 evals=evals,
                 ray_params=ray_params,
                 _checkpoint=checkpoint,
+                _additional_results=current_results,
                 **kwargs)
             break
         except RayActorError:
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -473,7 +473,7 @@ def _train(params: Dict,
                     if isinstance(item, Callable):
                         item()
                     elif isinstance(item, _Checkpoint):
-                        _checkpoint.__dict__ = item.__dict__
+                        _checkpoint.__dict__.update(item.__dict__)
                     else:
                         callback_returns[actor_rank].append(item)
             ready, not_ready = ray.wait(not_ready, timeout=0)
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -291,6 +291,11 @@ class RayXGBoostActor:
 
         local_params = params.copy()
 
+        if "xgb_model" in kwargs:
+            if isinstance(kwargs["xgb_model"], bytes):
+                # bytearray type gets lost in remote actor call
+                kwargs["xgb_model"] = bytearray(kwargs["xgb_model"])
+
         if "nthread" not in local_params:
             if num_threads > 0:
                 local_params["num_threads"] = num_threads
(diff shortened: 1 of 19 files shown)


xgboost_ray/main.py
```diff
@@ -257,16 +245,12 @@ def __init__(self,
...
-                                self.rank)
-
     @property
     def _save_checkpoint_callback(self):
         def callback(env):
-            if env.iteration % self.checkpoint_frequency == 0:
-                env.model.save_model(self.checkpoint_file)
+            if self.rank == 0 and \
+               env.iteration % self.checkpoint_frequency == 0:
+                put_queue(_Checkpoint(env.iteration, pickle.dumps(env.model)))
```
richardliaw, line 253: should we use something more efficient or put it into the object store?

richardliaw: Does it also work on GPUs?

xgboost_ray/tests/utils.py
```diff
@@ -59,3 +59,17 @@ def create_parquet_in_tempdir(filename: str,
         num_classes=num_classes,
         num_partitions=num_partitions)
     return temp_dir, path
+
+
+def flatten_obj(obj: Union[List, Dict], keys=None, base=None):
+    keys = keys if keys is not None else []
```
richardliaw, line 65: nit: keys = keys or []

xgboost_ray/main.py
```diff
@@ -291,6 +275,11 @@ def train(self, rabit_args: List[str], params: Dict[str, Any],
 
         local_params = params.copy()
 
+        if "xgb_model" in kwargs:
+            if isinstance(kwargs["xgb_model"], bytes):
+                # bytearray type gets lost in remote actor call
+                kwargs["xgb_model"] = bytearray(kwargs["xgb_model"])
+
```
richardliaw, line 282: is this an issue/bug?

xgboost_ray/main.py
```diff
@@ -257,16 +245,12 @@ def __init__(self,
...
         _set_omp_num_threads()
 
-    @property
-    def checkpoint_file(self) -> Optional[str]:
-        return _checkpoint_file(self.checkpoint_path, self.checkpoint_prefix,
-                                self.rank)
-
     @property
     def _save_checkpoint_callback(self):
         def callback(env):
```
richardliaw, line 250: Do you know why the xgboost internal callbacks use a Class-based callback?

https://github.com/dmlc/xgboost/blob/927c316aeb0a948e9635c1ffe54d5e6ed25f6797/python-package/xgboost/callback.py#L668-L677

xgboost_ray/main.py
```diff
@@ -496,6 +472,8 @@ def _train(params: Dict,
                     (actor_rank, item) = queue.get()
                     if isinstance(item, Callable):
                         item()
+                    elif isinstance(item, _Checkpoint):
+                        _checkpoint.__dict__ = item.__dict__
```
richardliaw, line 476: maybe instead, do `_checkpoint.__dict__.update(item.__dict__)`

xgboost_ray/main.py
```diff
@@ -421,6 +382,8 @@ def _train(params: Dict,
            *args,
            evals=(),
            ray_params: RayParams,
+           _checkpoint: _Checkpoint,
+           _additional_results: Dict,
```
richardliaw, line 386: hmm, what does this do?

xgboost_ray/main.py
krfricke: This seems to be a new change, here is the legacy callback handler: https://github.com/dmlc/xgboost/blob/927c316aeb0a948e9635c1ffe54d5e6ed25f6797/python-package/xgboost/callback.py#L682

It will be added in the next release (1.3.0, current stable release 1.2.x)

It should be easy to refactor our callbacks in the future, but I'd like to keep them in the legacy format to support older installations of xgboost. We should put this on our backlog though and possibly update in (late) Q1/2021

xgboost_ray/main.py
krfricke: Basically we're passing a checkpoint and results reference to be able to record (and return) the contents even in the case of failure. Since we re-raise the actor error in the try/catch block in `_train()`, this seemed like the best way to support keeping checkpoint and results states, and since `_train()` is an internal function it should be fine to use this API here. 
I'm open to suggestions here, though.

xgboost_ray/main.py
krfricke: It should work on GPUs, I'll try this though.

For the object store, should we `ray.put` the model into the object store and pass the reference via the queue?
Also, it would be great if you could explain why the current approach is inefficient, so that I gather some intuition about this. Thanks!

richardliaw: Yes! so from what i understand, the `model` is essentially a wrapper around some numerical datatype (array). Pickle serialization of numerical datatypes is often inefficient. Rather, the Plasma object store has much better suited for transferring numerical datatypes, so that's why `ray.put` is better! 

See: https://arrow.apache.org/blog/2017/10/15/fast-python-serialization-with-ray-and-arrow/

xgboost_ray/main.py
richardliaw: Cool sounds good!

richardliaw (approved):
Looks great!

xgboost_ray/main.py
krfricke: I think so. I'll check this tomorrow and write a small reproducible script, possibly open an issue on ray core

xgboost_ray/main.py
krfricke: Got it, thanks! I'll address this in a follow-up PR though to unblock the tune checkpointing

krfricke: Actually, I'll look into this before merging, it might be related to the error we currently see on the latest release

xgboost_ray/main.py
amogkam: Got it. So just for my understanding, `additional_results` and `num_boost_rounds` are not correctly being handled during failure in the current implementation correct? Without this PR, when a training run has to restart, each actor loads its local checkpoint, but it starts again from iteration 0 and previous `additional_results` are not being saved, is that right?

amogkam (approved):
LGTM! Just left one question to help with my understanding.

xgboost_ray/main.py
krfricke: Yes, that is exactly right. Without this PR the restarted actors train for `num_boost_rounds` again and forfeit any additional results they collected before.

krfricke:
Thanks, I'll look into the fail in the ray release sometime today/tomorrow before merging.

PR #28, opened 2020-12-15
Placement Group Support

amogkam:
Depends on https://github.com/ray-project/xgboost_ray/pull/29.

Adds placement group for xgboost training workers. 

Logic is as follows:
* If not using Tune:
    * If elastic_training == True
        * Don't use placement groups (Requires placement group resizing https://github.com/ray-project/ray/issues/12562)
    * If elastic_training == False
        * Use placement group with SPREAD strategy
* If using Tune:
    * If elastic_training == True
        * Raise Error. Tune does not support elastic training.
    * If elastic_training == False
        * Use placement group with PACK strategy


amogkam:
@rkooo567 if you get a chance, do you think you can give this a quick look as well to see if I'm using placement groups correctly?

xgboost_ray/main.py
```diff
@@ -411,13 +413,15 @@ def _create_actor(rank: int,
...
+                  placement_group: Optional[PlacementGroup] = None,
                   queue: Optional[Queue] = None,
                   checkpoint_frequency: int = 5) -> ActorHandle:
 
     return RayXGBoostActor.options(
         num_cpus=num_cpus_per_actor,
         num_gpus=num_gpus_per_actor,
-        resources=resources_per_actor).remote(
+        resources=resources_per_actor,
+        placement_group=placement_group).remote(
```
amogkam, line 424: @rkooo567 do we still have to pass in `num_cpus`, `num_gpus`, `resources` here if we are also passing in the placement group?

xgboost_ray/main.py
```diff
@@ -869,17 +914,20 @@ def train(params: Dict,
                 time.sleep(5)
                 queue.shutdown()
                 stop_event.shutdown()
+                #remove_placement_group(pg)
```
amogkam, line 917: Removing the placement group here causes the alive actors to die as well. I think we'd need to use placement group resizing here instead.

amogkam: Or I think another workaround we can do here is to disable the placement group-level fault tolerance.

amogkam:
Fyi @krfricke I’m planning to work on this more tomorrow. Can you take a look today and give some early feedback?

xgboost_ray/main.py
```diff
@@ -512,18 +517,51 @@ def _shutdown(actors: List[ActorHandle],
...
+    return pg
 
 
 def _train(params: Dict,
            dtrain: RayDMatrix,
            *args,
            evals=(),
            ray_params: RayParams,
+           cpus_per_actor: int,
+           gpus_per_actor: int,
```
krfricke, line 558: Should we still read those from `ray_params` (and not pass them here?)

xgboost_ray/main.py
```diff
@@ -816,8 +852,14 @@ def train(params: Dict,
...
-    queue = Queue()  # Queue actor
-    stop_event = Event()  # Stop event actor
+
+    # Create the Queue and Event actors.
+    queue, stop_event = _create_communication_processes()
+
+    pg = _create_placement_group(cpus_per_actor, gpus_per_actor,
+                                 ray_params.resources_per_actor,
+                                 ray_params.num_actors)
+
```
krfricke, line 862: The main question I have here is if we should always use placement groups even if we're not using Tune. Does this make a difference in practice? If not using Tune we would probably want to spread actors as much as possible.

xgboost_ray/util.py
```diff
@@ -37,8 +39,12 @@ def shutdown(self):
             ray.kill(self.actor)
         self.actor = None
 
-
+# TODO: Propogate this into RayQueue.
 class Queue(RayQueue):
+    def __init__(self, maxsize: int = 0, actor_options: Dict = {}) -> None:
+        self.maxsize = maxsize
+        self.actor = _QueueActor.options(**actor_options).remote(self.maxsize)
+
```
krfricke, line 47: I think we could actually consolidate both functionalities into one, e.g. create something like a `RayCommunication` actor class which has an `asyncio.Event` and `asyncio.Queue` and provides the interface for both.

krfricke: (Maybe out of scope for this PR though)

krfricke (commented):
Draft looks good so far. Placement group resizing is probably the greatest blocker right now.

richardliaw:
I'd advocate for only implementing placement group support for non-resizeable workloads. 

Currently (and for the near future), Tune is going to be the primary benefit for placement group support on workers, and these worker groups will be static in size.

amogkam:
> Currently (and for the near future), Tune is going to be the primary benefit for placement group support on workers, and these worker groups will be static in size.

@richardliaw can you elaborate on this? Does this mean that Tune will support Pack strategy on workers, and once that's implemented we would remove it from xgboost_ray?

richardliaw:
No, I'm just saying that if you use Tune, you'll need placement groups, but don't need elastic support for placement groups. If you don't use Tune, maybe we can get away without using placement groups entirely (for workers).

amogkam:
Why can’t Tune be used with scale-down elastic training? In that case the worker group size won’t be static.

richardliaw:
There is currently no support for that (maybe later, but out of scope for the next 4 weeks).

richardliaw:
When a node dies, Tune currently does not know which trials resources are lost. The current model for global resource changes is for the trial whose resources have changed to be relaunched with appropriate resources.

amogkam:
But scale down elastic training is being xgboost level, right? So as long as the Trainable is still alive, from Tunes perspective there is no failure since it’s being handled in xgboost_ray.train.

richardliaw:
Tune keeps track of global cluster resources (total and allocated). Without retriggering execution of the trial, Tune will not know that one trial has reduced its resource usage. This puts the resource accounting at an inconsistent state, and can lead to a variety of problems.

This is again, a limitation of Tune resource accounting model. We are revisiting it in this sprint, but before we actually make changes to Tune, anything that doesn't work with its existing form will not work.

xgboost_ray/main.py
```diff
@@ -512,18 +517,51 @@ def _shutdown(actors: List[ActorHandle],
...
+    node_ip = ray.services.get_node_ip_address()
+    # Have to explicitly set num_cpus to 0.
+    placement_option = {"num_cpus": 0, "resources": {f"node:{node_ip}": 0.01}}
+    queue = Queue(actor_options=placement_option)  # Queue actor
+    stop_event = Event(actor_options=placement_option)  # Stop event actor
+    return queue, stop_event
+
+def _create_placement_group(cpus_per_actor, gpus_per_actor,
+                            resources_per_actor, num_actors):
+    resources_per_bundle = {"CPU": cpus_per_actor, "GPU": gpus_per_actor}
```
rkooo567, line 534: Can you add a check that cpu per actor & gpu per actors are not both 0? Placement group won't work in that case

xgboost_ray/main.py
```diff
@@ -512,18 +517,51 @@ def _shutdown(actors: List[ActorHandle],
...
+    # Create placement group for training worker colocation.
+    bundles = [{**resources_per_bundle, **extra_resources_per_bundle} for _ in range(num_actors)]
+    pg = placement_group(bundles, strategy="PACK")
+    # Wait for placement group to get created.
+    logger.debug("Waiting for placement group to start.")
+    ready = ray.wait([pg.ready()], timeout=100)
+    if ready:
+        logger.debug("Placement group has started.")
+    else:
+        raise TimeoutError("Placement group creation timed out. Make sure "
```
rkooo567, line 546: Maybe you can also print available resources and required resources?

rkooo567 (commented):
LGTM for the placement group!

xgboost_ray/main.py
amogkam: Added!

xgboost_ray/main.py
amogkam: Good point- added!

xgboost_ray/util.py
amogkam: Yes that's a great idea- we should do that. I raised an issue to keep track of this.

xgboost_ray/main.py
amogkam: Right, this is one of the assumptions made in the doc that if not using Tune, the cluster is fully utilized (all CPUs are being used by Xgboost Actors). In this case, it doesn't matter whether we use Pack, Spread, or no placement groups at all.

But you're right if we don't make that assumption, we would want to use Spread strategy. I added that in here since it was pretty straightforward.

xgboost_ray/main.py
amogkam: These have gone through automatic detection already at this point, so they might not be the same as what's in ray_params.

xgboost_ray/main.py
```diff
@@ -855,7 +922,7 @@ def train(params: Dict,
                     f"Sleeping for 10 seconds for cleanup.")
                 start_again = True
 
-            if tries + 1 <= max_actor_restarts:
+            elif tries + 1 <= max_actor_restarts:
```
amogkam, line 925: I believe this should be an elif @krfricke? `max_actor_restarts` should be ignored if elastic training is True if I'm not mistaken.

xgboost_ray/main.py
```diff
@@ -800,7 +824,37 @@ def train(params: Dict,
...
+        if "tree_method" in params and params["tree_method"].startswith("gpu"):
+            gpus_per_actor = 1
+
+    # Automatically set cpus_per_actor if left at the default value
+    # Will be set to the number of cluster CPUs divided by the number of
+    # actors, bounded by the maximum number of CPUs across actors nodes.
+    if cpus_per_actor <= 0:
+        cluster_cpus = _ray_get_cluster_cpus() or 1
+        cpus_per_actor = min(
+            int(_get_min_node_cpus() or 1),
```
amogkam, line 845: Btw I think this should be `_get_min_node_cpus`, not `_get_max_node_cpus`. If we have a cluster with 3 nodes: 1 with 1 CPU, and 2 with 3 CPUS each, and we have num_actors=4, how many cpus should we have per actor?

Previously this would give 2 cpus per actor, but that's currently not possible with this setup. We should be bounded by the _minimum_ number of CPUs not the maximum, if I'm understanding this correctly.

amogkam:
Now that #29 is merged, this is ready for review @krfricke @richardliaw

xgboost_ray/main.py
richardliaw: Hmm, usually we should still have a max_actor_restart for elastic training. This usually prevents infinite loops for some obscure error.

xgboost_ray/main.py
```diff
@@ -836,11 +893,14 @@ def train(params: Dict,
                 *args,
                 evals=evals,
                 ray_params=ray_params,
+                cpus_per_actor=cpus_per_actor,
+                gpus_per_actor=gpus_per_actor,
```
richardliaw, line 897: maybe instead we can pass this in through a `actor_options` kwarg instead of 3 args?

xgboost_ray/main.py
```diff
@@ -495,23 +499,54 @@ def _get_actor_alive_status(actors: List[ActorHandle],
...
+        resources_per_actor
+    # Create placement group for training worker colocation.
+    bundles = [{
+        **resources_per_bundle,
+        **extra_resources_per_bundle
+    } for _ in range(num_actors)]
+    pg = placement_group(bundles, strategy=strategy)
+    # Wait for placement group to get created.
+    logger.debug("Waiting for placement group to start.")
+    ready, _ = ray.wait([pg.ready()], timeout=100)
```
richardliaw, line 539: can you make the timeout configurable (i.e via env var)?

xgboost_ray/main.py
richardliaw: I think so, yeah

xgboost_ray/main.py
richardliaw: Hmm, ideally you end up with 3 actors, but I think you'll want to do 3 CPUs each per actor (resulting in 2 created actors). 

The reason is because xgboost is largely memory bound, and doubling 2 actors onto 1 machine will be less desirable (increases the likelihood of OOM).

xgboost_ray/main.py
```diff
@@ -810,7 +826,37 @@ def train(params: Dict,
...
+        if "tree_method" in params and params["tree_method"].startswith("gpu"):
+            gpus_per_actor = 1
+
+    # Automatically set cpus_per_actor if left at the default value
+    # Will be set to the number of cluster CPUs divided by the number of
+    # actors, bounded by the minimum number of CPUs across actors nodes.
+    if cpus_per_actor <= 0:
+        cluster_cpus = _ray_get_cluster_cpus() or 1
+        cpus_per_actor = min(
+            int(_get_min_node_cpus() or 1),
```
richardliaw, line 847: Can we move all the code of 831 to 852 to a separate function like:

```
cpus_per_actor, gpus_per_actor = _validate_resources(...)
```

richardliaw (commented):
Just a couple comments. It's very close!

xgboost_ray/main.py
amogkam: Good idea- I added PLACEMENT_GROUP_TIMEOUT_S env var for this.

xgboost_ray/main.py
amogkam: I moved into a separate function

xgboost_ray/main.py
amogkam: Right, but we currently don't automatically configure the number of actors, only the number of _cpus/gpus_ per actor. So if the the user passes in `num_actors=4`, we will always create 4 actors even if it's undesirable.

Also, the example cluster setup I gave was incorrect, it should be 4 nodes in total, 2 with 1 CPU each, 2 with 3 CPUs each. 

In master, autodetecting CPUs would set `num_cpus_per_actor` to 2 with this setup:
`_get_max_node_cpus()` would give 3,
`int(cluster_cpus // num_actors)` == `int(8 // 4)` which would give 2,
and we would take the min of these to get 2 cpus per actor.
But having 4 actors with 2 CPUs each is not able to be scheduled with this cluster setup.

Instead we want to have `get_min_node_cpus` which would return 1. So we would have 4 actors with 1 CPU each which can be scheduled.

xgboost_ray/main.py
amogkam: I think there's 2 points here: whether `max_actor_restarts` should have an impact with elastic training and whether line 925 should be an if or elif.

For the first point, no actors are restarted with elastic training, so what would `max_actor_restarts` be used for? Also we say in the docstring that `max_actor_restarts` is ignored during elastic training (which intuitively is what I would expect as a user). But perhaps some other behavior makes more sense here.

But regardless of the first point, I think line 925 should be an elif. Functionally the only change it makes is not logging an incorrect logger.warning message.

xgboost_ray/main.py
richardliaw: I see; that makes sense.

xgboost_ray/main.py
richardliaw: Thanks for clearing that up, I missed that context. The elif looks right then.

xgboost_ray/main.py
```diff
@@ -39,6 +41,9 @@
 from xgboost_ray.session import init_session, put_queue, \
     set_session_queue
 
+# How long to wait for placement group creation before failing.
+PLACEMENT_GROUP_TIMEOUT_S = int(os.getenv("PLACEMENT_GROUP_TIMEOUT_S", 100))
```
krfricke, line 45: Let's document this somewhere

xgboost_ray/main.py
```diff
@@ -406,18 +411,42 @@ def predict(self, model: xgb.Booster, data: RayDMatrix, **kwargs):
         return predictions
 
 
+def _autodetect_resources(ray_params: Union[None, RayParams, Dict] = None,
+                          use_tree_method: bool = False):
```
krfricke, line 415:
```suggestion
                          use_tree_method: bool = False) -> Tuple[int, int]:
```

xgboost_ray/main.py
```diff
@@ -406,18 +411,42 @@ def predict(self, model: xgb.Booster, data: RayDMatrix, **kwargs):
...
+        if use_tree_method:
+            gpus_per_actor = 1
+
+    # Automatically set cpus_per_actor if left at the default value
+    # Will be set to the number of cluster CPUs divided by the number of
+    # actors, bounded by the minimum number of CPUs across actors nodes.
+    if cpus_per_actor <= 0:
+        cluster_cpus = _ray_get_cluster_cpus() or 1
+        cpus_per_actor = min(
+            int(_get_min_node_cpus() or 1),
```
krfricke, line 431: Changing this from max to min leads to quite different behavior, but I guess we can assume that people with heterogeneous cluster setups might set this manually anyway.

xgboost_ray/main.py
krfricke: It should be an `elif`. I added that somewhere, but it might have been reverted in some merge action before

xgboost_ray/main.py
```diff
@@ -826,7 +884,18 @@ def train(params: Dict,
...
     actors = [None] * ray_params.num_actors  # All active actors
+
+    # Create the Queue and Event actors.
     queue, stop_event = _create_communication_processes()
+
+    if not ray_params.elastic_training:
+        strategy = "PACK" if added_tune_callback else "SPREAD"
+        pg = _create_placement_group(cpus_per_actor, gpus_per_actor,
+                                     ray_params.resources_per_actor,
+                                     ray_params.num_actors, strategy)
```
krfricke, line 895: The main question I have here is if we always want to create placement groups. I'm not familiar enough with them to know the benefits and drawbacks. I get that we would want them with Tune, but without Tune and without elastic training, what is the benefit of using placement groups? Does SPREAD give us any benefits? Do placement groups have any significant overhead?

krfricke (commented):
This looks really good. I have a general question for placement groups, and I would like to see this in a medium scale release test. I'll run this tomorrow after benchmarking to make sure that it works.

xgboost_ray/main.py
krfricke: (So I'm fine with this change)

xgboost_ray/main.py
amogkam: @richardliaw @rkooo567 please feel free to add more info/correct me if I'm wrong here:

One of the benefits of placement groups is **atomic resource reservation** which is useful on a non-autoscaling cluster. With placement groups, if all the required resources cannot be reserved at the same time, then execution will fail (unless autoscaling is enabled). Without placement groups, some of the training workers will be scheduled, and others will not, causing execution to hang I believe. Note that if we are on an autoscaling cluster, both cases should trigger autoscaling, so there should not be a difference between the two.

As for the actual strategy, PACK vs. SPREAD vs. no placement groups should not make a difference if the entire Ray cluster is fully utilized. But if that's not the case, then SPREAD would improve fault tolerance since it minimizes the number of workers that fail when a node goes down (https://docs.ray.io/en/master/auto_examples/placement-group.html#improve-fault-tolerance). However, this comes at a cost of increased inter-node communication. So SPREAD prioritizes fault tolerance over reducing network communication. PACK does the complete opposite and prioritizes reducing network communication over fault tolerance. Not using placement groups does not make any prioritization, and the placement behavior is not deterministic.

As for additional overheads, I don't know the particulars here, but I'm sure at the very least there is some accounting overhead. But I believe that this overhead only occurs for scheduling decisions when actors are created. Once all of the training workers have been placed, actual training times should not increase.

amogkam: Overall, I think for the non-tune case SPREAD is what we want here since it will improve fault tolerance. But, I'm also OK if you think we should discuss this further, and we can disable placement groups for the non-tune case. It's only a 1 line change to toggle this.

xgboost_ray/main.py
amogkam: Just to be clear, for homogenous clusters this does not make a difference since min==max. But for heterogenous clusters I believe using `max` is actually incorrect behavior.

xgboost_ray/main.py
amogkam: We can also add an environment variable to toggle this, which will probably make it easier to run benchmarking with/without placement groups.

amogkam: I added an environment variable to toggle this for your benchmarking tomorrow.

xgboost_ray/main.py
amogkam: I changed it to just 1 additional arg for the placement group by setting the new cpus/gpus per actor back to ray_params.

xgboost_ray/main.py
krfricke: Awesome. Thanks for the thorough explanation. It sounds like placement groups are generally desired default behavior. I'll try it out tomorrow. If it works I don't have anything else to add and we should be ready to merge.

xgboost_ray/main.py
krfricke: min is definitely the safe choice. Generally,  max makes sure that at least one node can schedule an actor, min makes sure that each node can schedule actors. I had max in as a sanity check, to make sure that the number of requested CPUs is never larger than the largest node size, just as an upper bound. For scheduling this was desirable in the case where we have a small head node and large workers (e.g. for GPU training, where I used this setup).

Generally we should just stress to set num cpus per actor manually. We might even want to log a warning if we connect to an existing cluster and num cpus per actor is not set. Let's discuss this sometime!

xgboost_ray/main.py
amogkam: I'd actually prefer to leave it as is so it's clear as to what all of the arguments are. Plus this is an internal function so it should be fine to add these extra args. @krfricke any thoughts here?

xgboost_ray/main.py
amogkam: I added all of this to the README!

README.md
```diff
@@ -160,6 +160,29 @@ setting this explicitly.
...
+`xgboost_ray` leverages Ray's Placement Group API (https://docs.ray.io/en/master/placement-group.html)
+to implement placement strategies for better fault tolerance. 
+
+By default, a SPREAD strategy is used for training, which attempts to spread all of the training workers
+across the nodes in a cluster on a best-effort basis. This improves fault tolerance since it minimizes the 
+number of worker failures when a node goes down, but comes at a cost of increased inter-node communication
+To disable this strategy, set the `USE_SPREAD_STRATEGY` environment variable to 0. If disabled, no
+particular placement strategy will be used.
+
+Note that this strategy is used only when `elastic_training` is not used. If `elastic_trainin` is set to `True`,
```
amogkam, line 174:
```suggestion
Note that this strategy is used only when `elastic_training` is not used. If `elastic_training` is set to `True`,
```

krfricke:
These are some results for running with tune:

```
For IP 172.31.24.143 got trial IDs ['a159f_00000', 'a159f_00000', 'a159f_00000', 'a159f_00000', 'a159f_00001', 'a159f_00001', 'a159f_00001', 'a159f_00001']
For IP 172.31.26.125 got trial IDs ['a159f_00002']
For IP 172.31.23.54 got trial IDs ['a159f_00002', 'a159f_00002', 'a159f_00002', 'a159f_00003', 'a159f_00003', 'a159f_00003']
For IP 172.31.26.111 got trial IDs ['a159f_00003']
```

```
For IP 172.31.26.111 got trial IDs ['c961f_00000', 'c961f_00000', 'c961f_00000', 'c961f_00003', 'c961f_00003', 'c961f_00003']
For IP 172.31.23.54 got trial IDs ['c961f_00000', 'c961f_00001', 'c961f_00003']
For IP 172.31.24.143 got trial IDs ['c961f_00001', 'c961f_00001', 'c961f_00001']
For IP 172.31.26.125 got trial IDs ['c961f_00002', 'c961f_00002', 'c961f_00002', 'c961f_00002']
```

This does seem like PACK is working, right?

krfricke (approved):

PR #31, opened 2020-12-17
Add support for modin dataframes

krfricke:
Currently only in central data loading mode. Integration with distributed mode would require more access to private modin functions - we can look into this at another time.

This PR also adds (and fixes) compatibility tests for the latest ray release when some dependencies are not installed. E.g. if Tune or Modin are not installed, certain tests and examples should still run.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -117,8 +117,8 @@ jobs:
     - name: Uninstall unavailable dependencies
       # Disables modin and Ray Tune (via tabulate)
       run: |
-        python -m pip uninstall modin || true
-        python -m pip uninstall tabulate || true
+        python -m pip uninstall -y modin
+        python -m pip uninstall -y tabulate
     - name: Install package
       run: |
         python -m pip install -e .
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -95,3 +95,45 @@ jobs:
         echo "running simple_tune.py" && python simple_tune.py --smoke-test
         echo "running train_on_test_data.py" && python train_on_test_data.py --smoke-test
       # for f in *.py; do echo "running $f" && python "$f" || exit 1 ; done
+
+  test_linux_compat:
+    # Test compatibility when some optional libraries are missing
+    # Test runs on latest ray release
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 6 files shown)


xgboost_ray/matrix.py
```diff
@@ -17,13 +17,18 @@
...
 except ImportError:
+    MLDataset = Unavailable
 
-    class MLDataset:
-        pass
-
+try:
+    import modin  # noqa: F401
+    MODIN_INSTALLED = True
+except ImportError:
```
amogkam, line 30: Do you have to set `modin = None` here?

.github/workflows/test.yaml
```diff
@@ -95,3 +95,45 @@ jobs:
         echo "running simple_tune.py" && python simple_tune.py --smoke-test
         echo "running train_on_test_data.py" && python train_on_test_data.py --smoke-test
       # for f in *.py; do echo "running $f" && python "$f" || exit 1 ; done
+
+  test_linux_compat:
```
amogkam, line 99: Nice!

amogkam (approved):
Looks good to me! 

Just 2 questions:
1) Do we have to update the RayDMatrix docstring since data can now be an mldataset or moding dataframe?
2) Should we add an example for data loading via moding (maybe in a separate PR)?

We should also create an issue for adding modin support for distributed data loading just to keep track of it.

xgboost_ray/matrix.py
```diff
@@ -306,7 +343,8 @@ def load_data(self,
 
         if self.label is not None and not isinstance(self.label, str) and \
            not (isinstance(self.data, pd.DataFrame) and
-                isinstance(self.label, pd.Series)):
+                isinstance(self.label, pd.Series)) and \
+           not (_is_modin_df(self.data) and _is_modin_series(self.label)):
```
amogkam, line 347: Maybe move this condition logic to a separate function? Something like `is_valid_label`?

xgboost_ray/matrix.py
krfricke: No, this is just to check if we can import it (`F401` is the warning that a module is imported but unused). We cannot actually import the real required modin modules (DataFrame, Series) here as it would initialize ray then.

xgboost_ray/matrix.py
krfricke: Since we only need this once at this place I'll leave it here for the time being. If this grows another time, we should move this

krfricke:
I updated the docstrings. As for the example, we could do that, but we might want to wait until we have support for distributed loading. Right now it would be just a duplication of passing a pandas dataframe/series, and even the data distribution would be the same.

PR #33, opened 2020-12-18
Update README for peak memory usage suggestions

krfricke:
There's not much to do to reduce peak memory usage. Thus we should just document things to consider and best practices.


richardliaw (approved):
looks good; merge when done?

PR #34, opened 2020-12-21
Only replace callbacks if necessary (removes unnecessary Tune warning)

krfricke:
Since we inherit from the Ray Tune-specific callbacks, the callbacks were replaced and a warning was thrown even if it wasn't necessary.


amogkam (approved):
Good catch!

PR #46, opened 2021-01-27
Add petastorm integration

krfricke:


xgboost_ray/matrix.py
```diff
@@ -424,7 +458,9 @@ def load_data(self,
 
         invalid_data = False
         if isinstance(self.data, str):
-            if os.path.isdir(self.data):
+            if self.filetype is RayFileType.PETASTORM:
```
richardliaw, line 461: nit - ==

xgboost_ray/tests/test_matrix.py
```diff
@@ -110,6 +110,39 @@ def testFromModinDfString(self):
         in_df["label"] = self.y
         self._testMatrixCreation(in_df, "label")
 
+    def testFromPetastormParquetString(self):
```
richardliaw, line 113: hmm since petastorm is automatically enabled, would some previous tests be implicitly changed to leverage petastorm when we should instead leave it to test the non-petastorm codepath?

I didn't actually check the tests - but just a thought :)

richardliaw (approved):
looks good! left a couple nits

xgboost_ray/tests/test_matrix.py
krfricke: They won't, as petastorm requires an explicit `file://` prefix - and we don't have that in the other tests :-)

xgboost_ray/matrix.py
krfricke, line 461:
```suggestion
            if self.filetype == RayFileType.PETASTORM:
```

PR #47, opened 2021-01-27
Remove kill callbacks from colocation tests

krfricke:
This will hopefully deflake the colocation tests

### DIFF
--- a/xgboost_ray/tests/conftest.py
+++ b/xgboost_ray/tests/conftest.py
@@ -57,8 +57,9 @@ def _ray_start_cluster(**kwargs):
 
 
 # This fixture will start a cluster with empty nodes.
-@pytest.fixture
+@pytest.fixture(scope="class")
 def ray_start_cluster(request):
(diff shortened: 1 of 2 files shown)


amogkam:
Hey @krfricke if these tests are flaky with the kill callback, then doesn't that mean there's a problem with the functionality? These should be passing even with worker failures, right?

krfricke:
So the error that comes up is usually that actors cannot be scheduled because resources aren't ready, and this might actually be a core issue. It also seems to come up in the `latest` release and not in the `nightly`.

Generally I'd argue that we want to keep the tests atomic - i.e. if we check colocation, just check colocation. If we check fault tolerance, just check fault tolerance. I'm happy to add another test that combines both but we might want to do this in a separate PR.

krfricke:
(This would probably be just a PR for using Ray Tune + Fault tolerance)

amogkam (approved):
Got it, makes sense. The tune fault tolerance test sounds like a good idea.

PR #52, opened 2021-02-04
Limit number of parallel GHA runs, print env info in tests

krfricke:
Unfortunately GHA runs are sometimes aborted due to memory limits. Reducing the number of concurrent trials increases our test time to 2x, but at least these errors stop coming up.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,12 +29,14 @@ jobs:
     timeout-minutes: 12
     strategy:
       matrix:
-        python-version: [3.7, 3.8]
+        python-version: [3.6.9, 3.7, 3.8]
         include:
+          - python-version: 3.6.9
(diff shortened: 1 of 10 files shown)


krfricke:
Well, it seems it doesn't solve the problem. Consider this a draft PR, I'll look into how to fix this later

amogkam (approved):
Interesting- great catch!

amogkam:
Though it still looks like 2 of the jobs are failing.

krfricke:
Yes :-( Maybe we can move this to buildkite?

krfricke:
I don't think this is related to that, the tests were flaky before that change. I mean, we can try, I'll create a revert PR just in case

krfricke:
Doesn't seem to be possible from within GitHub. I might try it manually tomorrow

amogkam:
@krfricke should we merge this?

krfricke:
Yeah let's do it, though it doesn't solve the problem completely. But it is a good start in any case. Thanks for reminding me!

PR #54, opened 2021-02-10
Tune reporting

krfricke:
Currently, we are reporting results from every actor. This leads to a double reports and a blown up iteration count. Instead, we should just report results from the 0 rank actor.


amogkam (approved):
This is great, thanks!

PR #55, opened 2021-02-11
Add get_tune_resources helper method

amogkam:
Will need to change after https://github.com/ray-project/ray/pull/13906 is merged.

### DIFF
--- a/xgboost_ray/tests/test_tune.py
+++ b/xgboost_ray/tests/test_tune.py
@@ -63,21 +63,24 @@ class XGBoostRayTuneTest(unittest.TestCase):
         ray.shutdown()
         shutil.rmtree(self.experiment_dir)
 
-    # noinspection PyTypeChecker
-    def testNumIters(self):
-        ray_params = RayParams(cpus_per_actor=1, num_actors=1)
+    def numItersCheck(self, num_actors):
+        ray_params = RayParams(cpus_per_actor=1, num_actors=num_actors)
         analysis = tune.run(
             self.train_func(ray_params),
             config=self.params,
-            resources_per_trial={
-                "cpu": 1,
-                "extra_cpu": 1
-            },
+            resources_per_trial=ray_params.get_tune_resources(),
             num_samples=2)
 
         self.assertTrue(
             all(analysis.results_df["training_iteration"] ==
                 analysis.results_df["config.num_boost_round"]))
+
+    # noinspection PyTypeChecker
+    def testNumItersSingleActor(self):
+        self.numItersCheck(num_actors=1)
+
+    def testNumItersMultipleActor(self):
+        self.numItersCheck(num_actors=2)
 
     def testElasticFails(self):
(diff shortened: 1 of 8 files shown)


xgboost_ray/tune.py
```diff
@@ -180,3 +191,41 @@ def _try_add_tune_callback(kwargs: Dict):
...
+                        resources_per_actor: Optional[Dict]):
+    """Returns object to use for ``resources_per_trial`` with Ray Tune."""
+    if TUNE_INSTALLED:
+        if TUNE_1_2:
+            if resources_per_actor is not None:
+                raise RuntimeError("Using xgboost_ray with custom actor "
+                                   "resources with Ray Tune version 1.2 or "
+                                   "prior is not supported. Please install "
+                                   "the nightly Ray wheels to upgrade your "
+                                   "Tune version.")
```
krfricke, line 207: As discussed, enable extra custom resources

xgboost_ray/tune.py
```diff
@@ -38,11 +40,20 @@ def is_session_enabled():
...
+    TUNE_1_1 = False
 
-if TUNE_LEGACY and TUNE_INSTALLED:
+# Todo(amogkam): Remove after Ray 1.3 release.
+try:
+    from ray.tune.resources import PlacementGroupFactory
+
+    TUNE_1_2 = False
+except ImportError:
+    TUNE_1_2 = True
```
krfricke, line 53: Rename

krfricke (changes requested):

xgboost_ray/tune.py
amogkam: Done!

xgboost_ray/tune.py
amogkam: Done!

xgboost_ray/tests/test_tune.py
```diff
@@ -65,14 +65,11 @@ def tearDown(self):
 
     # noinspection PyTypeChecker
     def testNumIters(self):
-        ray_params = RayParams(cpus_per_actor=1, num_actors=2)
+        ray_params = RayParams(cpus_per_actor=1, num_actors=1)
```
krfricke, line 68: Can we leave this at 2? This was actually an error before (the num iters was incorrect because we reported multiple times from different actor ranks). Keeping this will make sure we don't regress

krfricke (approved):

krfricke:
Should be fine, thanks. Just revert the num trials to 2 if possible please!

xgboost_ray/tests/test_tune.py
amogkam: Oops that's my bad. I had changed this for some debugging and forgot to switch back. Thanks for pointing this out.

amogkam, line 68:
```suggestion
        ray_params = RayParams(cpus_per_actor=1, num_actors=2)
```

PR #66, opened 2021-03-12
Re-factor data loading structure

krfricke:
The current structure for data loading was a bit cluttered, and it was unclear how to add new data sources or extend existing ones. By moving data sources into their own sub package, we can set this up for future integrations.

### DIFF
--- a/xgboost_ray/data_sources/data_source.py
+++ b/xgboost_ray/data_sources/data_source.py
@@ -18,10 +18,34 @@ class DataSource:
     @staticmethod
     def is_data_type(data: Any,
                      filetype: Optional[RayFileType] = None) -> bool:
+        """Check if the supplied data matches this data source.
+
+        Args:
+            data (Any): Data set
+            filetype (Optional[RayFileType]): Optional RayFileType. Some
+                data sources might require that this is explicitly set
+                (e.g. if multiple sources can read CSV files).
+
+        Returns:
+            Boolean indicating if this data source belongs to/is compatible
+                with the data.
+        """
         return False
 
     @staticmethod
     def get_filetype(data: Any) -> Optional[RayFileType]:
+        """Method to help guessing the filetype.
+
+        Should return None if the supplied data type (usually a filename)
+        is not covered by this data source, otherwise the filetype should
+        be returned.
+
+        Args:
+            data (Any): Data set
+
+        Returns:
+            RayFileType or None.
+        """
         return None
 
     @staticmethod
@@ -29,10 +53,25 @@ class DataSource:
                   ignore: Optional[Sequence[str]] = None,
                   indices: Optional[Sequence[int]] = None,
                   **kwargs) -> pd.DataFrame:
+        """
+        Load data into a pandas dataframe.
+
+        Ignore specific columns, and optionally select specific indices.
+
+        Args:
+            data (Any): Input data
+            ignore (Optional[Sequence[str]]): Column names to ignore
+            indices (Optional[Sequence[int]]): Indices to select. What an
+                index indicates depends on the data source.
+
+        Returns:
+            Pandas DataFrame.
+        """
         raise NotImplementedError
 
     @staticmethod
     def convert_to_series(data: Any) -> pd.Series:
+        """Convert data from the data source type to a pandas series"""
         if isinstance(data, pd.DataFrame):
             return pd.Series(data.squeeze())
 
@@ -44,6 +83,10 @@ class DataSource:
     @classmethod
     def get_column(cls, data: pd.DataFrame,
                    column: Any) -> Tuple[pd.Series, Optional[str]]:
+        """Helper method wrapping around convert to series.
+
+        This method should usually not be overwritten.
+        """
         if isinstance(column, str):
             return data[column], column
         elif column is not None:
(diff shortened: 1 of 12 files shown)


xgboost_ray/data_sources/data_source.py
```diff
@@ -0,0 +1,99 @@
...
+class RayFileType(Enum):
+    """Enum for different file types (used for overrides)."""
+    CSV = 1
+    PARQUET = 2
+    PETASTORM = 3
+
+
+class DataSource:
+    supports_central_loading = True
+    supports_distributed_loading = False
```
richardliaw, line 16: Consider documenting this class

xgboost_ray/data_sources/ml_dataset.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, Optional, Sequence, List
+
+import pandas as pd
+from ray.util.data import MLDataset as MLDatasetType
+from xgboost_ray.data_sources.data_source import DataSource, RayFileType
+
+
+class MLDataset(DataSource):
```
richardliaw, line 8: consider documenting

xgboost_ray/data_sources/modin.py
```diff
@@ -0,0 +1,52 @@
...
+import pandas as pd
+
+try:
+    import modin  # noqa: F401
+    MODIN_INSTALLED = True
+except ImportError:
+    MODIN_INSTALLED = False
+
+
+class Modin(DataSource):
```
richardliaw, line 14: consider documenting

xgboost_ray/data_sources/petastorm.py
```diff
@@ -0,0 +1,55 @@
...
+    @staticmethod
+    def get_filetype(data: Any) -> Optional[RayFileType]:
+        if not PETASTORM_INSTALLED:
+            return None
+
+        if not isinstance(data, List):
+            data = [data]
+
+        def _is_compatible(url: str):
+            return url.endswith(".parquet") and (url.startswith("s3://")
```
richardliaw, line 32: btw doesnt petastorm also support other filetypes? (out of scope)

xgboost_ray/data_sources/modin.py
```diff
@@ -0,0 +1,52 @@
...
+                     filetype: Optional[RayFileType] = None) -> bool:
+        if not MODIN_INSTALLED:
+            return False
+        from modin.pandas import DataFrame as ModinDataFrame, \
+            Series as ModinSeries
+
+        return isinstance(data, (ModinDataFrame, ModinSeries))
+
+    @staticmethod
+    def load_data(
```
richardliaw, line 26: do you need to check modin installation for all functions?

xgboost_ray/data_sources/data_source.py
```diff
@@ -0,0 +1,99 @@
...
+    @staticmethod
+    def is_data_type(data: Any,
+                     filetype: Optional[RayFileType] = None) -> bool:
+        """Check if the supplied data matches this data source.
+
+        Args:
+            data (Any): Data set
+            filetype (Optional[RayFileType]): Optional RayFileType. Some
+                data sources might require that this is explicitly set
+                (e.g. if multiple sources can read CSV files).
```
richardliaw, line 27:
```suggestion
        Args:
            data (Any): Dataset.
            filetype (Optional[RayFileType]): RayFileType of the provided
                dataset. Some DataSource implementations might require 
                that this is explicitly set (e.g. if multiple sources can 
                read CSV files).
```

xgboost_ray/data_sources/data_source.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Returns:
+            Boolean indicating if this data source belongs to/is compatible
+                with the data.
+        """
+        return False
+
+    @staticmethod
+    def get_filetype(data: Any) -> Optional[RayFileType]:
+        """Method to help guessing the filetype.
```
richardliaw, line 37:
```suggestion
        """Method to help infer the filetype.
```

xgboost_ray/data_sources/data_source.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Should return None if the supplied data type (usually a filename)
+        is not covered by this data source, otherwise the filetype should
+        be returned.
```
richardliaw, line 41:
```suggestion
        Returns None if the supplied data type (usually a filename)
        is not covered by this data source, otherwise the filetype
        is returned.
```

richardliaw: I would stick with a more imperative wording here even though this is an abstract class

richardliaw (approved):
looks good!

xgboost_ray/data_sources/petastorm.py
krfricke: It seems to be mainly used for reading parquet, but let's follow up on this in a separate issue

xgboost_ray/data_sources/modin.py
krfricke: No - if the data source is incompatible it will not be selected when trying to load data:

```
            if source.is_data_type(self.data, self.filetype):
                data_source = source
                break
```

krfricke: But I guess it can't hurt to add an `_assert_modin_installed()`

PR #69, opened 2021-03-17
Fix RayDeviceQuantileDMatrix

### LINKED ISSUE #65: Issue on Device Quantile DMatrix.
Hi team, it was mentioned during our offline chat that in some cases the Device Quantile DMatrix doesn't work.  It would be great if there's a description and a reproducible example for the issue so that we can try to fix it.

krfricke:
RayDeviceQuantileDMatrix became incompatible with recent changes. This PR makes sure it works again.

Please note that we currently have no general support to load from files into GPU memory directly, avoiding CPU memory. Thus the RayDeviceQuantileDMatrix works with conversion (or when cupy arrays are provided directly). In the future we might want to introduce interfaces for loading from files into GPU memory.

Closes #65

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -215,8 +215,6 @@ def _get_dmatrix(data: RayDMatrix, param: Dict) -> xgb.DMatrix:
                 "feature_types": data.feature_types,
                 "missing": data.missing,
             }
-            if not isinstance(data, xgb.DeviceQuantileDMatrix):
-                pass
             param.update(dm_param)
             it = RayDataIter(**param)
             matrix = xgb.DeviceQuantileDMatrix(it, **dm_param)
@@ -432,7 +430,10 @@ class RayXGBoostActor:
             if num_threads > 0:
                 local_params["num_threads"] = num_threads
             else:
-                local_params["nthread"] = ray.utils.get_num_cpus()
+                try:
+                    local_params["nthread"] = ray._private.utils.get_num_cpus()
+                except AttributeError:
+                    local_params["nthread"] = ray.utils.get_num_cpus()
 
         if dtrain not in self._data:
             self.load_data(dtrain)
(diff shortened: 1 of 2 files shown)


trivialfis:
> Please note that we currently have no general support to load from files into GPU memory directly.

That will require direct access to external memory from GPU.  I think it's best for a 3-party library to carry out this kind of task.  WDYT?

xgboost_ray/main.py
```diff
@@ -432,7 +442,10 @@ def train(self, rabit_args: List[str], params: Dict[str, Any],
             if num_threads > 0:
                 local_params["num_threads"] = num_threads
             else:
-                local_params["nthread"] = ray.utils.get_num_cpus()
+                try:
+                    local_params["nthread"] = ray._private.utils.get_num_cpus()
+                except AttributeError:
+                    local_params["nthread"] = ray.utils.get_num_cpus()
```
richardliaw, line 448: what's the purpose of this? maybe we can use `get_resource_ids()` instead of tapping the private call?

krfricke: Ah, the resource IDs is actually much better as we will only assign CPUs we actually requested resources for. Thanks, updated.

krfricke:
Yes @trivialfis I agree loading from files directly into GPU memory is a task for some other library.

What we could do in this integration though is that currently when we're loading from multiple CSV/Parquet files, we're constructing pandas dataframes in CPU memory and concatenate them. When creating device quantile dmatrices, we convert the (usually full) dataframe to cupy arrays in the iterator and pass them to the device matrix.

Instead we could not create pandas dataframes, but create cupy arrays to begin with instead. We would then not hold a copy of the full data in CPU memory. For construction, data will still flow through CPU memory but only until the cupy array is built.

Does that make sense?

richardliaw (approved):

trivialfis:
Thanks for the explanation, yeah that make sense.

PR #72, opened 2021-03-19
Fault tolerance benchmarking utilities + Release test refactor + e2e GPU test

krfricke:
Also includes a manual workflow to trigger e2e distributed GPU tests


.github/workflows/gpu.yaml
```diff
@@ -0,0 +1,34 @@
+name: GPU on manual trigger
```
amogkam, line 1: Just curious, how would you trigger this? Perhaps adding some small instructions for how to run the release tests would help here.

amogkam (approved):

.github/workflows/gpu.yaml
krfricke: I think there's supposed to be a button under the "Actions" tab here on GitHub after merging. Guess we'll find out!

Issue #80, opened 2021-04-12
Do you provide a technical architecture diagram？


Matrix-World:
In the README.md, I only saw some usage and case introductions, but I don't know how xgboost and ray are combined. I tried to read the source code and drew the architecture diagram myself, but it was not particularly clear. So can you provide one? Can illustrate the process of an app running.
Looking forward your reply.

krfricke:
Hi @Matrix-World, we're preparing a short paper/poster for XGBoost-Ray anyway, so we might end up doing something like this eventually. Is there something that you're particularly interested in and find hard to understand?

Matrix-World:
> Hi @Matrix-World, we're preparing a short paper/poster for XGBoost-Ray anyway, so we might end up doing something like this eventually. Is there something that you're particularly interested in and find hard to understand?

yes, When will you publish your paper?
There is a problem that I find very strange. I successfully ran the py script(https://github.com/ray-project/xgboost_ray/blob/master/examples/train_on_test_data.py) and saw the usage of CPU and Plasma in the Ray Web.
But when I set gpus_per_actor=1 and num_actor=4, I did not see the gpu in use, and no error was reported, and the code ran successfully. Can you answer it?
--------below is my code---------
import argparse
import os
import shutil
import time

from xgboost_ray import train, RayDMatrix, RayParams
from xgboost_ray.tests.utils import create_parquet_in_tempdir
def main(fname, num_actors, cpus_per_actor):
    dtrain = RayDMatrix(
        os.path.abspath(fname), label="labels", ignore=["partition"])

    config = {
        "tree_method": "hist",
        "eval_metric": ["logloss", "error"],
    }

    evals_result = {}

    start = time.time()
    bst = train(
        config,
        dtrain,
        evals_result=evals_result,
        ray_params=RayParams(max_actor_restarts=1,
                             num_actors=num_actors,
                             gpus_per_actor=1,
                             cpus_per_actor=cpus_per_actor),
        num_boost_round=50,
        evals=[(dtrain, "train")])

    taken = time.time() - start
    print(f"Train Time Taken: {taken:.2f} seconds")

    bst.save_model("test_data_for_xgboost.xgb")
    print("Final training error: {:.4f}".format(
        evals_result["train"]["error"][-1]))


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--smoke-test",
        action="store_true",
        default=True,
        help="Finish quickly for testing")
[...]

PR #82, opened 2021-04-14
Show warning if gpus are assigned but tree method is incompatible

krfricke:
Closes issue discussed in #80

Should we raise an error instead?


amogkam (approved):
I think warning should be fine for now.

PR #89, opened 2021-04-21
[Private APIs] Switch to `ray.worker.get_resource_ids`

ijrsvt:
* `get_resource_ids` are being moved out of the top level `ray` API in this [Ray PR](https://github.com/ray-project/ray/pull/15440).

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -184,7 +184,7 @@ class RabitContext:
 
 def _ray_get_actor_cpus():
     # Get through resource IDs
-    resource_ids = ray.get_resource_ids()
+    resource_ids = ray.worker.get_resource_ids()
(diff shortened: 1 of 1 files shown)


krfricke:
Thanks @ijrsvt. We'll need to keep this backwards compatible with the latest Ray release, so it might be best to move the `get_resource_ids` to a separate function until the change is reflected in the next release. I'm happy to update your PR with that.

ijrsvt:
@krfricke using the private API works on *all* versions of Ray. This is where the function is actually defined :) 

It was just imported to the top level `ray.get_resource_ids`

krfricke:
Ah sorry, I mixed that up - thanks!

krfricke (approved):

PR #90, opened 2021-04-23
Compatibility for xgboost 0.90

krfricke:
cc @mmui

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,5 +10,5 @@ setup(
     "distributed computing framework Ray.",
     url="https://github.com/ray-project/xgboost_ray",
     install_requires=[
-        "xgboost>=1.3", "ray", "numpy>=1.16,<1.20", "pandas", "pyarrow"
+        "xgboost", "ray", "numpy>=1.16,<1.20", "pandas", "pyarrow"
     ])
--- a/setup.py
+++ b/setup.py
@@ -10,5 +10,5 @@ setup(
     "distributed computing framework Ray.",
     url="https://github.com/ray-project/xgboost_ray",
     install_requires=[
-        "xgboost", "ray", "numpy>=1.16,<1.20", "pandas", "pyarrow"
+        "xgboost>=0.90", "ray", "numpy>=1.16,<1.20", "pandas", "pyarrow"
     ])
--- a/xgboost_ray/compat/__init__.py
+++ b/xgboost_ray/compat/__init__.py
@@ -10,7 +10,7 @@ except ImportError:
         f"(version {xgb.__version__}). While we try to support "
         f"older XGBoost versions, please note that this library is only "
         f"fully tested and supported for XGBoost >= 1.4. Please consider "
-        f"upgrading your XGBoost version (`pip install -u xgboost`).")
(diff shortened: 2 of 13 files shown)


setup.py
```diff
@@ -10,5 +10,5 @@
     "distributed computing framework Ray.",
     url="https://github.com/ray-project/xgboost_ray",
     install_requires=[
-        "xgboost>=1.3", "ray", "numpy>=1.16,<1.20", "pandas", "pyarrow"
+        "xgboost", "ray", "numpy>=1.16,<1.20", "pandas", "pyarrow"
```
richardliaw, line 13: >=0.90?

xgboost_ray/compat/tracker.py
```diff
@@ -0,0 +1,361 @@
+# flake8: noqa
+# Copied from https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/tracker.py
```
richardliaw, line 2: Do you need to add some licensing notes here?

setup.py
krfricke: I guess we could add `>=0.90` - this branch is still compatible with xgboost >= 1.3 (and probably with any version between those, but I haven't tried)

xgboost_ray/compat/tracker.py
krfricke: Right, this is Apache 2.0 so probably yeah. I'll quickly check how to attribute this correctly

setup.py
richardliaw: yeah that'd be great

xgboost_ray/compat/tracker.py
krfricke: Ok since our own license is also Apache 2.0 we shouldn't run into any problems.

I still added a license header to this file and linked to the correct (hash-identified) file source and license file on the xgboost repo.

krfricke: (pushed but github seems to need some time to update this PR)

xgboost_ray/compat/__init__.py
```diff
@@ -0,0 +1,51 @@
...
+try:
+    from xgboost.callback import TrainingCallback
+    LEGACY_CALLBACK = False
+except ImportError:
+    logging.warning(
+        f"You are using `xgboost_ray` with a legacy XGBoost version "
+        f"(version {xgb.__version__}). While we try to support "
+        f"older XGBoost versions, please note that this library is only "
+        f"fully tested and supported for XGBoost >= 1.4. Please consider "
+        f"upgrading your XGBoost version (`pip install -U xgboost`).")
```
richardliaw, line 13: i wonder if this will get printed on every single Ray worker?

krfricke: I don't remember seeing that in my runs but I can check again later to make sure.

richardliaw (approved):

xgboost_ray/compat/__init__.py
krfricke: Turns out you were right - I moved the warning to the `predict()` and `train()` methods, so they will only be printed once.

PR #99, opened 2021-05-12
Add support for dask dataframes

### LINKED ISSUE #92: Support Distributed Dask dataframes
As we're seeing large improvements on the Ray-Dask front, I think it makes sense that the Xgboost Ray support a direct ingest from Dask. 

cc @worldveil @clarkzinzow

krfricke:
Closes #92

Note that the locality-aware scheduling can currently not be tested reliably. This is due to 2 reasons:

1) We cannot guarantee that dask partitions end up living on a specific node (to set an initial state for redistribution)
2) We cannot guarantee that tasks that determine the dask partition node are co-scheduled with the respective partition.

Once the `ray.state.objects()` API or something similar comes back, the second option should be easy to enable, and the first option will be easy to confirm.

### DIFF
--- a/examples/simple_dask.py
+++ b/examples/simple_dask.py
@@ -5,8 +5,6 @@ import pandas as pd
 
 import ray
 
-import dask.dataframe as dd
-
 from xgboost_ray import RayDMatrix, train, RayParams
 from xgboost_ray.data_sources.dask import DASK_INSTALLED
 
@@ -15,6 +13,12 @@ def main(cpus_per_actor, num_actors):
     if not DASK_INSTALLED:
         print("Dask is not installed. Install with `pip install dask`")
         return
+
+    # Local import so the installation check comes first
+    import dask
+    import dask.dataframe as dd
+    from ray.util.dask import ray_dask_get
+    dask.config.set(scheduler=ray_dask_get)
 
     # Generate dataset
     x = np.repeat(range(8), 16).reshape((32, 4))
--- a/examples/simple_dask.py
+++ b/examples/simple_dask.py
@@ -1,0 +1,91 @@
+import argparse
+
+import numpy as np
+import pandas as pd
+
+import ray
+
+import dask.dataframe as dd
+
+from xgboost_ray import RayDMatrix, train, RayParams
+from xgboost_ray.data_sources.dask import DASK_INSTALLED
+
+
+def main(cpus_per_actor, num_actors):
+    if not DASK_INSTALLED:
+        print("Dask is not installed. Install with `pip install dask`")
+        return
+
+    # Generate dataset
+    x = np.repeat(range(8), 16).reshape((32, 4))
+    # Even numbers --> 0, odd numbers --> 1
+    y = np.tile(np.repeat(range(2), 4), 4)
+
+    # Flip some bits to reduce max accuracy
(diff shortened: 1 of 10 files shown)


krfricke:
Currently it seems that the problem lies with 1) - `ray memory` shows that each partition lives on the head node after calling `persist()`. I'll try to find a solution to that.

examples/simple_dask.py
```diff
@@ -0,0 +1,91 @@
...
+
+import dask.dataframe as dd
+
+from xgboost_ray import RayDMatrix, train, RayParams
+from xgboost_ray.data_sources.dask import DASK_INSTALLED
+
+
+def main(cpus_per_actor, num_actors):
+    if not DASK_INSTALLED:
+        print("Dask is not installed. Install with `pip install dask`")
```
richardliaw, line 16: should you also enable the dask-ray scheduler in this example?

xgboost_ray/data_sources/dask.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if indices is not None and len(indices) > 0 and isinstance(
+                indices[0], Tuple):
+            # We got a list of partition IDs belonging to Dask partitions
+            return dd.concat(
+                [data.partitions[i] for (i, ) in indices]).compute()
+
+        # Dask does not support iloc() for row selection, so we have to
+        # compute a local pandas dataframe first
+        local_df = data.compute()
```
richardliaw, line 72: will this materialize the full distributed dataframe?

xgboost_ray/data_sources/dask.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if isinstance(obj_ref, ray.ObjectRef):
+            node_id = partition_locations[pid]
+            node = node_ids_to_node.get(node_id, {})
+            ip = node.get("NodeManagerAddress", "_no_ip")
+        else:
+            ip = "_no_ip"
+
+        # Pass tuples here (integers can be misinterpreted as row numbers)
+        ip_to_parts[ip].append((pid, ))
```
richardliaw, line 151: hmm, why would it be misinterpreted?

richardliaw (approved):
Generally looks great! The main question i have is about some of the implementation details.

xgboost_ray/data_sources/dask.py
krfricke: Yes. However, this part of the code concerns the local loading - i.e. loading of either a partition or centralized loading, where access to the whole dataframe is assumed anyway. This is the part where we end up with a pandas dataframe.

xgboost_ray/data_sources/dask.py
krfricke: In `load_data()`, we are doing this check:

```

        if indices is not None and len(indices) > 0 and isinstance(
                indices[0], Tuple):
```

Usually, numerical indices would indicate the row numbers a worker should load. This is also valid for Dask (e.g. for centralized loading). However, in this case the numbers should indicate the partition numbers. To not confuse these with row indices, we use a different type. IMO tuples are a good choice because they are immutable and should produce very few overhead.

Note that in Modin we use Ray ObjectIDs instead, but we don't have these for Dask at this point in the code.

PR #100, opened 2021-05-12
Only one remote actor should return its Booster

krfricke:
Currently each actor returns the trained Booster object.

This is inefficient:

1. All Booster objects are the same
2. Transferring large Boosters over slow networks can lead to much overhead
3. We call `ray.get()` three times on each of the training futures. Although results should be cached, this doesn't always seem to be the case

Instead, only the first live training actor should return its Booster object. 

Also, we want to minimize the number of calls to `ray.get()`.

This should fix an issue that came up training with 150 workers where each individual result fetch would take about 2 seconds to complete.

cc @mmui

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -924,12 +924,12 @@ def _train(params: Dict,
     ]
     training_futures = [
         actor.train.remote(
+            rabit_args,
+            i == 0,  # return_bst
+            params,
+            dtrain,
+            evals,
             *args,
-            rabit_args=rabit_args,
-            return_bst=i == 0,
-            params=params,
-            dtrain=dtrain,
-            evals=evals,
             **kwargs) for i, actor in enumerate(live_actors)
     ]
 
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -459,8 +459,9 @@ class RayXGBoostActor:
 
         self._distributed_callbacks.after_data_loading(self, data)
 
-    def train(self, rabit_args: List[str], params: Dict[str, Any],
-              dtrain: RayDMatrix, evals: Tuple[RayDMatrix, str], *args,
+    def train(self, rabit_args: List[str], return_bst: bool,
+              params: Dict[str, Any], dtrain: RayDMatrix,
+              evals: Tuple[RayDMatrix, str], *args,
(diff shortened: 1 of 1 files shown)


krfricke:
Before we merge this I would like to confirm that this actually resolves the issue. I.e. it could also be a large evaluation results object.

krfricke:
Confirmed with a local change that this does indeed solve the issue in the specific run.

xgboost_ray/main.py
```diff
@@ -914,9 +919,12 @@ def handle_actor_failure(actor_id):
     _training_state.training_started_at = time.time()
 
     # Trigger the train function
+    live_actors = [
+        actor for actor in _training_state.actors if actor is not None
+    ]
     training_futures = [
-        actor.train.remote(rabit_args, params, dtrain, evals, *args, **kwargs)
-        for actor in _training_state.actors if actor is not None
+        actor.train.remote(rabit_args, i == 0, params, dtrain, evals, *args,
```
heyitsmui, line 926: could one of the live_actors go away during the training process or fail to return the bst? if so, should we include a retry here to fetch the bst from the next available actor?

heyitsmui:
Thanks for the PR, this would speed up post-processing for jobs with large models and/or a large number of actors!

xgboost_ray/main.py
krfricke: No, if the designated return actor dies we will restart the `_train()` function in any case and the next live actor will become the new designated return actor

xgboost_ray/main.py
```diff
@@ -459,8 +459,9 @@ def load_data(self, data: RayDMatrix):
 
         self._distributed_callbacks.after_data_loading(self, data)
 
-    def train(self, rabit_args: List[str], params: Dict[str, Any],
-              dtrain: RayDMatrix, evals: Tuple[RayDMatrix, str], *args,
+    def train(self, rabit_args: List[str], return_bst: bool,
+              params: Dict[str, Any], dtrain: RayDMatrix,
+              evals: Tuple[RayDMatrix, str], *args,
```
richardliaw, line 464: can you make these named arguments?

richardliaw (approved):
Looks good, one nit

PR #103, opened 2021-05-13
Fix Tune PG import

amogkam:


amogkam:
Btw @krfricke do you think we should just remove resources dict support (pre Ray 1.3), and only do PlacementGroupFactory?

krfricke (approved):

krfricke:
Let's remove it after 1.4 release

PR #108, opened 2021-05-25
Add examples with Ray Client to CI

amogkam:


xgboost_ray/examples/simple_tune.py
```diff
@@ -38,6 +39,13 @@ def train_breast_cancer(config, ray_params):
         evals_result["eval"]["error"][-1]))
 
 
+def load_best_model(checkpoint_dir):
+    """Loads the XGBoost model."""
+    best_bst = xgb.Booster()
+    best_bst.load_model(os.path.join(checkpoint_dir, "tuned.xgb"))
+    return best_bst
+
+
```
krfricke, line 48: Can we consolidate this with the client check and put into a proper xgboost_ray util?

xgboost_ray/util.py
```diff
@@ -145,3 +145,30 @@ def is_ready(self):
...
+        raise ValueError("Cannot found the node dictionary for current node.")
+
+
+def force_on_current_node(task_or_actor):
+    """Given a task or actor, place it on the current node.
+
+    If using Ray Client, the current node is the client server node.
+    """
+    node_resource_key = get_current_node_resource_key()
+    options = {"resources": {node_resource_key: 0.01}}
```
krfricke, line 173: Does this do a deep update?

krfricke (changes requested):
Can we add a symlink to xgboost_ray/examples on the top level?

xgboost_ray/examples/simple_tune.py
amogkam: Done!

xgboost_ray/util.py
amogkam: No this doesn't do a deep update, but I don't think it should. If we want to force on current node then the task or actor should be placed on the current node, ignoring other custom resource requests. I added a comment in the docstring for this.

krfricke: (We don't use the utility to place training actors or alike)

krfricke (approved):

mirekphd:
@krfricke I think this commit made example files much harder to find... note that symlinks work only locally... on the web just use ordinary links;)

Empty folders (due to failing symlinks):
https://github.com/ray-project/xgboost_ray/blob/master/examples
https://github.com/ray-project/lightgbm_ray/blob/master/examples
https://github.com/ray-project/lightgbm_ray/blob/main/examples

Missing file (HTTP 404):
https://github.com/ray-project/xgboost_ray/blob/main/examples

vs. correct ones to link (not symlink) to:
https://github.com/ray-project/lightgbm_ray/tree/main/lightgbm_ray/examples
https://github.com/ray-project/xgboost_ray/tree/master/xgboost_ray/examples

PR #110, opened 2021-06-04
Fix issues with objective="multi:softprob"

Yard1:
`objective="multi:softprob"` returns a 2d array instead of a 1d array during prediction. This caused issues with `predict` as it expected a 1d array. This PR fixes this, as well as the handling of BATCH sharding mode during prediction (though I think it still fails on training).

The only change to the public API would be that the callback receives a pandas DataFrame instead of a Series, but I don't think it will cause many issues.

A test was added.

### DIFF
--- a/xgboost_ray/callback.py
+++ b/xgboost_ray/callback.py
@@ -45,8 +45,8 @@ class DistributedCallback(ABC):
     def before_predict(self, actor: "RayXGBoostActor", *args, **kwargs):
         pass
 
-    def after_predict(self, actor: "RayXGBoostActor", predictions: pd.Series,
-                      *args, **kwargs):
+    def after_predict(self, actor: "RayXGBoostActor",
+                      predictions: pd.DataFrame, *args, **kwargs):
         pass
 
 
@@ -81,8 +81,8 @@ class DistributedCallbackContainer:
         for callback in self.callbacks:
             callback.before_predict(actor, *args, **kwargs)
 
-    def after_predict(self, actor: "RayXGBoostActor", predictions: pd.Series,
-                      *args, **kwargs):
+    def after_predict(self, actor: "RayXGBoostActor",
+                      predictions: pd.DataFrame, *args, **kwargs):
         for callback in self.callbacks:
             callback.after_predict(actor, predictions, *args, **kwargs)
 
--- a/xgboost_ray/callback.py
+++ b/xgboost_ray/callback.py
@@ -1,5 +1,5 @@
 from abc import ABC
-from typing import Dict, Sequence, TYPE_CHECKING, Any
+from typing import Dict, Sequence, TYPE_CHECKING, Any, Union
 
 import os
 import pandas as pd
@@ -46,7 +46,8 @@ class DistributedCallback(ABC):
         pass
 
     def after_predict(self, actor: "RayXGBoostActor",
-                      predictions: pd.DataFrame, *args, **kwargs):
+                      predictions: Union[pd.Series, pd.DataFrame], *args,
+                      **kwargs):
         pass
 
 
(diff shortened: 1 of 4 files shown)


Yard1:
CC @richardliaw

xgboost_ray/matrix.py
```diff
@@ -896,19 +896,41 @@ def _get_sharding_indices(sharding: RayShardingMode, rank: int,
...
+        # need to be handled differently
+        if sharding == RayShardingMode.BATCH:
+            res = np.vstack(data)
+        elif sharding == RayShardingMode.INTERLEAVED:
+            # Sometimes the lengths are off by 1 for uneven divisions
+            min_len = min(len(d) for d in data)
+            # the number of classes will be constant, though
+            class_len = data[0].shape[1]
+            min_len_data = [d[0:min_len] for d in data]
+            res = np.hstack(min_len_data).reshape(
```
richardliaw, line 931: why do you hstack instead of vstack here?

Yard1: hstacking combined with reshape interleaves the rows (consistent with 1d array case)

xgboost_ray/callback.py
```diff
@@ -45,8 +45,8 @@ def after_train(self, actor: "RayXGBoostActor", result_dict: Dict, *args,
     def before_predict(self, actor: "RayXGBoostActor", *args, **kwargs):
         pass
 
-    def after_predict(self, actor: "RayXGBoostActor", predictions: pd.Series,
-                      *args, **kwargs):
+    def after_predict(self, actor: "RayXGBoostActor",
+                      predictions: pd.DataFrame, *args, **kwargs):
```
krfricke, line 49: Should we accept `Union[pd.Series, pd.DataFrame]` here?

xgboost_ray/main.py
```diff
@@ -578,8 +578,9 @@ def predict(self, model: xgb.Booster, data: RayDMatrix, **kwargs):
             self.load_data(data)
         local_data = self._data[data]
 
-        predictions = pd.Series(model.predict(local_data, **kwargs))
-        self._distributed_callbacks.after_predict(self, predictions)
+        predictions = model.predict(local_data, **kwargs)
+        self._distributed_callbacks.after_predict(self,
+                                                  pd.DataFrame(predictions))
```
krfricke, line 583: And here convert to Series/DataFrame based on model output?

xgboost_ray/matrix.py
```diff
@@ -896,19 +896,41 @@ def _get_sharding_indices(sharding: RayShardingMode, rank: int,
...
         raise ValueError(f"Invalid value for `sharding` parameter: "
                          f"{sharding}"
                          f"\nFIX THIS by passing any item of the "
                          f"`RayShardingMode` enum, for instance "
                          f"`RayShardingMode.BATCH`.")
+
+    # discard empty arrays that show up with BATCH
+    data = [d for d in data if len(d)]
+
+    if data[0].ndim == 1:
```
krfricke, line 909: We should probably check for the type here. Maybe also use `Iterable[Union[pd.Series, pd.DataFrame]]` for the type hint in the function signature.

krfricke (commented):
Generally looks good, just one question regarding conditionally passing Series/Dataframes

xgboost_ray/matrix.py
Yard1: With the conversion to pandas done just for the callback, the data here will always be composed of plain numpy arrays.

xgboost_ray/main.py
Yard1: I was thinking about it but I feel that it will actually be easier for users to know that the callback always receives a dataframe and not either one or the other. If they would like to have a series they can easily convert it.

krfricke (approved):

Issue #96, opened 2021-05-10
[feat req] Support Scikit-learn interface


richardliaw:
We should allow users to use `XGBoostClassifier` with XGBoostRay.

richardliaw:
(assign to @Yard1)

PR #111, opened 2021-06-07
Add a scikit-learn interface

Yard1:
Will close https://github.com/ray-project/xgboost_ray/issues/96 after it's finished.

Also includes https://github.com/ray-project/xgboost_ray/pull/110, so that should be merged first.

~~Issue: As this interface inherits from xgboost's sklearn interface, it may be hard if not impossible to make this interface compatible with all xgboost versions supported by xgboost-ray. We may have to consider setting an additional xgboost version requirement just for this interface in order to avoid issues.~~ After changes seems to support legacy XGBoost now.

`tests/test_sklearn.py` is copied almost verbatim from https://github.com/dmlc/xgboost/blob/a5c852660b1056204aa2e0cbfcd5b4ecfbf31adf/tests/python/test_with_sklearn.py in order to ensure 1:1 coverage, with minimal modifications. Some tests were disabled due to not being applicable for a distributed setting.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,7 +30,7 @@ jobs:
 
   test_linux_ray_master:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     strategy:
       matrix:
         python-version: [3.6.9, 3.7, 3.8]
@@ -68,7 +68,7 @@ jobs:
 
   test_linux_ray_release:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     strategy:
       matrix:
         python-version: [3.6.9, 3.7, 3.8]
@@ -101,7 +101,7 @@ jobs:
     # Test compatibility when some optional libraries are missing
     # Test runs on latest ray release
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     strategy:
       matrix:
         python-version: [3.6.9, 3.7, 3.8]
@@ -138,7 +138,7 @@ jobs:
   test_linux_cutting_edge:
     # Tests on cutting edge, i.e. latest Ray master, latest XGBoost master
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
     strategy:
       matrix:
         python-version: [3.6.9, 3.7, 3.8]
@@ -194,7 +194,7 @@ jobs:
   test_linux_xgboost_legacy:
     # Tests on XGBoost 0.90 and latest Ray release
     runs-on: ubuntu-latest
-    timeout-minutes: 18
+    timeout-minutes: 28
     strategy:
       matrix:
         python-version: [3.6.9]
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -221,7 +221,7 @@ jobs:
         ./xgboost_ray/tests/env_info.sh
     - name: Run tests
       run: |
-        bash ./run_ci_tests.sh
+        bash ./run_ci_tests.sh --legacy
     - name: Run examples
       run: |
         bash ./run_ci_examples.sh
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -221,7 +221,7 @@ jobs:
         ./xgboost_ray/tests/env_info.sh
     - name: Run tests
       run: |
-        bash ./run_ci_tests.sh --legacy
+        bash ./run_ci_tests.sh
     - name: Run examples
       run: |
         bash ./run_ci_examples.sh
--- a/README.md
+++ b/README.md
@@ -90,6 +90,81 @@ pred_ray = predict(bst, dpred, ray_params=RayParams(num_actors=2))
 print(pred_ray)
 ```
 
+### scikit-learn API
+
+XGBoost-Ray also features a scikit-learn API fully mirroring pure
+XGBoost scikit-learn API, providing a completely drop-in
+replacement. The following estimators are available:
+
+- `RayXGBClassifier`
+- `RayXGRegressor`
+- `RayXGBRFClassifier`
+- `RayXGBRFRegressor`
+- `RayXGBRanker`
+
+Example usage of `RayXGBClassifier`:
+
+```python
+from xgboost_ray import RayXGBClassifier, RayParams
+from sklearn.datasets import load_breast_cancer
+from sklearn.model_selection import train_test_split
+
+seed = 42
+
+X, y = load_breast_cancer(return_X_y=True)
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, train_size=0.25, random_state=42
+)
+
+clf = RayXGBClassifier(
+    n_jobs=4,  # In XGBoost-Ray, n_jobs sets the number of actors
+    random_state=seed
+)
+
+# scikit-learn API will automatically conver the data
+# to RayDMatrix format as needed
+
+clf.fit(X_train, y_train)
+
+pred_ray = clf.predict(X_test)
+print(pred_ray)
+
+pred_proba_ray = clf.predict_proba(X_test)
+print(pred_proba_ray)
+
+# It is also possible to pass a RayParams object
+# to fit/predict/predict_proba methods - will override
+# n_jobs set during initialization
+
+clf.fit(X_train, y_train, ray_params=RayParams(num_actors=2))
+
+pred_ray = clf.predict(X_test, ray_params=RayParams(num_actors=2))
+print(pred_ray)
+```
+
+Things to keep in mind:
+
+- `n_jobs` parameter controls the number of actors spawned.
+You can pass a `RayParams` object to the
+`fit`/`predict`/`predict_proba` methods as the `ray_params` argument 
+for greater control over resource allocation. Doing
+so will override the value of `n_jobs` with the value of
+`ray_params.num_actors` attribute. For more information, refer
+to the [Resources](#resources) section below.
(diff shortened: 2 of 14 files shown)


xgboost_ray/sklearn.py
```diff
@@ -0,0 +1,682 @@
...
+                sample_weight=sample_weight,
+                base_margin=base_margin,
+                feature_weights=feature_weights,
+                eval_set=eval_set,
+                sample_weight_eval_set=sample_weight_eval_set,
+                base_margin_eval_set=base_margin_eval_set,
+                eval_group=None,
+                eval_qid=None,
+                create_dmatrix=lambda **kwargs: RayDMatrix(**kwargs),
+            )
```
krfricke, line 167: This code block seems to be in the other regressors as well. Does it make sense to have a `RayScikitLearnBase` (like Dask does) that implements some of this logic for all regressors?

krfricke (commented):
Looking good so far! I know it's still WIP. Just as a general feedback, we will want to reduce code duplication as much as possible, so if we can abstract some logic to a base class or so that might be a good option here.

xgboost_ray/main.py
```diff
@@ -475,13 +475,14 @@ def train(self, rabit_args: List[str], return_bst: bool,
...
 
-        if "nthread" not in local_params:
+        if "nthread" not in local_params and "n_jobs" not in local_params:
             if num_threads > 0:
                 local_params["num_threads"] = num_threads
             else:
                 local_params["nthread"] = sum(
                     num
                     for _, num in ray.worker.get_resource_ids().get("CPU", []))
+                local_params["n_jobs"] = local_params["nthread"]
```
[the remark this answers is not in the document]
Yard1: n_jobs and nthreads are aliases of each other. The xgboost documentation says that new users should use the former. xgboost-dask overwrites both with the same value, so that's what I've done here too.

xgboost_ray/sklearn.py
Yard1: XGBoost sklearn API has some funky way to dynamically generate `get_params` that breaks when introducing new classes to the inheritance. I suppose that doing something similar to xgboost-dask sklearn interface could work here. My reasoning was that by inheriting from the estimators themselves we would avoid breaking any `isinstance` checks in user code, making this 100% drag-and-drop replace, but you are right in saying that a base class would provide more flexibility.

xgboost_ray/sklearn.py
```diff
@@ -0,0 +1,682 @@
...
+    ray_params (Union[None, RayParams, Dict]): Parameters to configure
+        Ray-specific behavior. See :class:`RayParams` for a list of valid
+        configuration parameters.
+    _remote (bool): Whether to run the driver process in a remote
+        function. This is enabled by default in Ray client mode.
+
+"""
+
+_N_JOBS_DOC_REPLACE = (
+    """    n_jobs : int
```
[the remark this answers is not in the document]
krfricke: Ah here I meant that we're documenting `n_jobs` twice

xgboost_ray/main.py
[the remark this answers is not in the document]
krfricke: Thanks for the explanation!

xgboost_ray/__init__.py
```diff
@@ -2,11 +2,18 @@
 from xgboost_ray.matrix import RayDMatrix, RayDeviceQuantileDMatrix,\
     RayFileType, RayShardingMode, \
     Data, combine_data
+# workaround for legacy xgboost==0.9.0
+try:
+    from xgboost_ray.sklearn import RayXGBClassifier, RayXGBRegressor, \
+        RayXGBRFClassifier, RayXGBRFRegressor, RayXGBRanker
+except ImportError:
+    pass
```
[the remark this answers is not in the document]
Yard1: I hadn't caught any issues with this approach.

.github/workflows/test.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
 
   test_linux_ray_master:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
```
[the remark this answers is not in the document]
Yard1: We could split sklearn tests to a separate job?

xgboost_ray/tests/test_sklearn.py
```diff
@@ -0,0 +1,1147 @@
+"""Copied almost verbatim from https://github.com/dmlc/xgboost/blob/a5c852660b1056204aa2e0cbfcd5b4ecfbf31adf/tests/python/test_with_sklearn.py
+in order to ensure 1:1 coverage, with minimal modifications.
+Some tests were disabled due to not being applicable for a
+distributed setting."""  # noqa: E501
```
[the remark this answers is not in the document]
Yard1: I had to modify this to conform to our unit test structure, so there won't be an easy way to diff the changes. We probably will need to review it manually. Will include the license.

Yard1:
I think it's only missing readme changes & and example at this point

README.md
```diff
@@ -90,6 +90,81 @@ pred_ray = predict(bst, dpred, ray_params=RayParams(num_actors=2))
...
+`ray_params.num_actors` attribute. For more information, refer
+to the [Resources](#resources) section below.
+- By default `n_jobs` is set to `1`, which means the training
+will **not** be distributed. Make sure to either set `n_jobs`
+to a higher value or pass a `RayParams` as outlined above in
+order to take advantage of XGBoost-Ray's functionality.
+- Aside from `ray_params`, `fit`/`predict`/`predict_proba`
+methods also posses the `_remote` argument.
+- After calling `fit`, additional evaluation results will
+be available under `additional_results_` attribute.
```
krfricke, line 161: Add: (e.g. training time, number of rows, callback results)

README.md
```diff
@@ -90,6 +90,81 @@ pred_ray = predict(bst, dpred, ray_params=RayParams(num_actors=2))
...
+for greater control over resource allocation. Doing
+so will override the value of `n_jobs` with the value of
+`ray_params.num_actors` attribute. For more information, refer
+to the [Resources](#resources) section below.
+- By default `n_jobs` is set to `1`, which means the training
+will **not** be distributed. Make sure to either set `n_jobs`
+to a higher value or pass a `RayParams` as outlined above in
+order to take advantage of XGBoost-Ray's functionality.
+- Aside from `ray_params`, `fit`/`predict`/`predict_proba`
+methods also posses the `_remote` argument.
```
krfricke, line 159: Let's leave this out. `_remote` is a private argument and shouldn't be exposed to users

xgboost_ray/sklearn.py
```diff
@@ -0,0 +1,901 @@
...
+            n_jobs: int) -> RayParams:
+        """Helper function to set num_actors in ray_params if not
+        set by the user"""
+        if ray_params is None:
+            if not n_jobs or n_jobs < 1:
+                n_jobs = 1
+            ray_params = RayParams(num_actors=n_jobs)
+        else:
+            warnings.warn("`ray_params` is not `None` and will override "
+                          "the `n_jobs` attribute.")
```
krfricke, line 239: IMO we should only throw a warning here if `n_jobs` is not `None`. `n_jobs` should be `Optional[int]` anyway, right?

xgboost_ray/sklearn.py
```diff
@@ -0,0 +1,901 @@
...
+        if XGBOOST_VERSION_TUPLE < (1, 4, 0):
+            warnings.warn(LEGACY_WARNING)
+        return f(*args, **kwargs)
+
+    return inner_f
+
+
+class RayXGBMixin:
+    """Mixin class to provide xgboost-ray functionality"""
+
```
krfricke, line 227: Should we define `additional_results_` here so that the attribute always exists?

krfricke (commented):
Almost there, just a couple of nits

xgboost_ray/sklearn.py
Yard1: This is the sklearn way of doing things. Attributes that are only set after calling fit are set in fit and suffixed with an underscore. Setting this in init will cause issues with cloning and sklearn in general

xgboost_ray/sklearn.py
Yard1: Wanted to make it clear to the user that the n_jobs will be replaced by ray_params should they pass the latter

README.md
Yard1: Will do

xgboost_ray/sklearn.py
krfricke: Right, but this is only important for the user if they set n_jobs themselves, right? Otherwise it seems like expected behavior and shouldn't raise a warning

Yard1: Oh, right, yeah. I'll add a check for that.

xgboost_ray/sklearn.py
```diff
@@ -0,0 +1,901 @@
...
+    def _ray_set_ray_params_n_jobs(
+            self, ray_params: Optional[Union[RayParams, dict]],
+            n_jobs: int) -> RayParams:
+        """Helper function to set num_actors in ray_params if not
+        set by the user"""
+        if ray_params is None:
+            if not n_jobs or n_jobs < 1:
+                n_jobs = 1
+            ray_params = RayParams(num_actors=n_jobs)
+        elif n_jobs:
```
krfricke, line 237: nit:

```suggestion
        elif n_jobs is not None:
```

And can we change the type hint from n_jobs to `Optional[int]` (which I think is accurate after looking at the sklearn code)?

Yard1: It is indeed more accurate

krfricke (approved):
Thanks!!

PR #113, opened 2021-06-09
Fail CI if tests do not pass

Yard1:
This PR makes the CI scripts exit with correct status if at least one test fails. Previously, no matter what the test results were, the CI always passed (aside from linting). This may uncover some bad tests/errors that were not noticed before.

The `run_ci_tests.sh` script was made so that all tests will run instead of exiting CI as soon as a pytest invocation fails in order to simplify debugging. Same could be done for examples but I don't think it's worth it.

### DIFF
--- a/run_ci_examples.sh
+++ b/run_ci_examples.sh
@@ -1,3 +1,5 @@
+set -e
+
 TUNE=1
 
 for i in "$@"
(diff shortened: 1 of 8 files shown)


krfricke (approved):
Thanks!

Issue #112, opened 2021-06-08
Decrease test times / move to buildkite


krfricke:
In order to keep development up to speed, we should consider speeding up CI tests or move them to buildkite

PR #115, opened 2021-06-16
fix RayParam resource population

kyle-chen-uber:
small fix about resource population

### DIFF
--- a/xgboost_ray/tune.py
+++ b/xgboost_ray/tune.py
@@ -206,7 +206,7 @@ def _get_tune_resources(num_actors: int, cpus_per_actor: int,
                 else resources_per_actor
             extra_custom_resources = {
                 k: v * num_actors
-                for k, v in resources_per_actor
+                for k, v in resources_per_actor.items()
             }
             return dict(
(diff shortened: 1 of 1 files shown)


amogkam (approved):
Thanks!

amogkam:
@krfricke do you know whats happening with the CI here? Should we increase the timeout? Or just keep re-trying

krfricke:
I'm not sure why this is failing out of a sudden (it passed previously). But yeah, probably just increasing to 40 minutes is fine.

Mentioning #112 here, which we should address soon.

krfricke:
#117 should fix the test timeouts. I'll then re-run tests here again, and merge afterwards.

krfricke:
Hm doesn't seem to be related to timeouts. It only fails on latest ray master, so there might be incompatible changes there.

Actually, the xgboost-ray/tune release are failing right now, too, so this is probably related.

krfricke:
The failing tests are due to https://github.com/ray-project/ray/issues/16507

This PR should be safe to merge

PR #117, opened 2021-06-17
Increase test timeout limits

krfricke:
Currently tests are failing due to timeouts.

For now, let's increase test times. In the future, let's make sure to address #112 soon.


krfricke:
Closing this as the bug is actually in https://github.com/ray-project/ray/issues/16507

PR #120, opened 2021-06-17
Separate out `_RayXGBoostActor`

Yard1:
Separate out `_RayXGBoostActor` class to enable inheritance in projects using xgboost-ray as a dependency

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -351,8 +351,7 @@ def _validate_ray_params(ray_params: Union[None, RayParams, dict]) \
     return ray_params
 
 
-@ray.remote
-class RayXGBoostActor:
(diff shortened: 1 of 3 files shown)


xgboost_ray/main.py
```diff
@@ -605,6 +604,11 @@ def predict(self, model: xgb.Booster, data: RayDMatrix, **kwargs):
         return predictions
 
 
+@ray.remote
+class RayXGBoostActor(_RayXGBoostActor):
```
richardliaw, line 608: maybe the docs should go here instead, since this one is more public facing?

Yard1: They will be inherited, right?

krfricke (approved):

Issue #116, opened 2021-06-17
Incompatibility with Modin 0.10.0 DataFrame 


matthewdeng:
**Problem:** 
After bumping to `modin 0.10.0`, `xgboost_ray` no longer recognizes `modin`'s `DataFrame`. Calling `train` on a `modin` `DataFrame` results in the following error:
```
2021-06-16 21:55:57,585	INFO services.py:1315 -- View the Ray dashboard at http://127.0.0.1:8265
UserWarning: Distributing <class 'NoneType'> object. This may take some time.
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    train({}, matrix)
  File "/Users/matt/anaconda3/envs/ray/lib/python3.8/site-packages/xgboost_ray/main.py", line 1196, in train
    dtrain.load_data(ray_params.num_actors)
  File "/Users/matt/anaconda3/envs/ray/lib/python3.8/site-packages/xgboost_ray/matrix.py", line 741, in load_data
    refs, self.n = self.loader.load_data(
  File "/Users/matt/anaconda3/envs/ray/lib/python3.8/site-packages/xgboost_ray/matrix.py", line 317, in load_data
    data_source = self.get_data_source()
  File "/Users/matt/anaconda3/envs/ray/lib/python3.8/site-packages/xgboost_ray/matrix.py", line 278, in get_data_source
    raise ValueError(
ValueError: Unknown data source type: <class 'modin.pandas.dataframe.DataFrame'> with FileType: None.
FIX THIS by passing a supported data type. Supported data types include pandas.DataFrame, pandas.Series, np.ndarray, and CSV/Parquet file paths. If you specify a file, path, consider passing the `filetype` argument to specify the type of the source. Use the `RayFileType` enum for that.
```

**Steps to Reproduce:**

1. Install `modin 0.10.0`: `pip install -U modin==0.10.0` 
    a. This also happens with the latest commit: `pip install -U git+https://github.com/modin-project/modin`
2. Run the following python script:
```python
import modin.pandas as pd
import ray
from xgboost_ray import RayDMatrix, train


ray.init()
df = pd.DataFrame()
matrix = RayDMatrix(df, label="label")
train({}, matrix)
```

**Reference:**
This issue does not happen with the previous version of `modin 0.9.1`.

1. Downgrade to `modin 0.9.1`: `pip uninstall -y modin && pip install modin==0.9.1`
[...]

PR #121, opened 2021-06-18
Support modin 0.10.0

amogkam:
Currently we are using string comparison to check for valid modin versions `MODIN_INSTALLED = modin.__version__ >= "0.9.0"`. Modin recently released 0.10.0 last week which fails this check. Thus, if modin 0.10 is installed, xgboost-ray thinks that modin is not installed. This means that the modin CI tests are currently being skipped, and also causing xgboost-ray to not recognize modin as a valid data source (https://github.com/ray-project/xgboost_ray/issues/116). Instead, we should do a more robust version check. Also, this PR tests the simple modin example with Ray Client now that 0.10 is released.

Closes https://github.com/ray-project/xgboost_ray/issues/116

### DIFF
--- a/xgboost_ray/data_sources/modin.py
+++ b/xgboost_ray/data_sources/modin.py
@@ -14,7 +14,8 @@ from xgboost_ray.data_sources.object_store import ObjectStore
 
 try:
     import modin  # noqa: F401
-    MODIN_INSTALLED = modin.__version__ >= "0.9.0"
+    from distutils.version import LooseVersion
+    MODIN_INSTALLED = LooseVersion(modin.__version__) >= LooseVersion("0.9.0")
 except (ImportError, AttributeError):
     MODIN_INSTALLED = False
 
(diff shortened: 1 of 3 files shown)


xgboost_ray/data_sources/modin.py
```diff
@@ -14,7 +14,8 @@
 
 try:
     import modin  # noqa: F401
-    MODIN_INSTALLED = modin.__version__ >= "0.9.0"
+    from distutils.version import LooseVersion
+    MODIN_INSTALLED = LooseVersion(modin.__version__) >= LooseVersion("0.9.0")
```
matthewdeng, line 18: Should we use the more recent [packaging.version.parse](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python) instead?

amogkam: Yeah I didn't want to use that since I believe that's an additional dependency that needs to be installed and would thus have to be packaged with xgboost-ray. `distutils` is a built in package. Also `LooseVersion` is being used in tensorflow (https://github.com/tensorflow/tensorflow/search?q=LooseVersion) so I think should be fine.

matthewdeng (approved):

krfricke (approved):
Thanks for the catch! Will merge once we fixed the tests and tests pass

PR #122, opened 2021-06-18
[rfc] Change default value of placement group from None to "default"

scv119:
Please read https://github.com/ray-project/ray/issues/16507 for the bug and context.
Here is a summary:
- After #[16437](https://github.com/ray-project/ray/pull/16437) the Actor placement behavior [has changed for ](https://github.com/ray-project/ray/pull/16437/files#diff-f67ab59de16332a2ea8df0fb7032ea69a0912c3bb52538649b4ed0c9fc21d9d2R636)`placement_group = None` versus `placement_group = "default"`; 
- Also in #[16437](https://github.com/ray-project/ray/pull/16437), by default the `placement_group` is set to `"default"` instead of `None` [when creating Ray Actors](https://github.com/ray-project/ray/pull/16437/files#diff-f67ab59de16332a2ea8df0fb7032ea69a0912c3bb52538649b4ed0c9fc21d9d2R500).
- However in when tune_small.py test runs,  RayXGBoostActor's`placement_group` was to `None`, which leads to allocation failure;
- So [this line sets ](https://github.com/ray-project/xgboost_ray/blob/6b9fb5e2cf68ec0e0039702bcd0bd8f362931279/xgboost_ray/main.py#L1241) the RayXGBoostActor with None placement group; verified changing it to "default" fixes the issue.

This is an attempt to fix it in xgboost_ray by substituting the default placement group from `None` to `default`. However this might change the behavior if xgboost_ray running with old ray version. Let me know how do we want to deal with this api behavior change. cc @ericl  @krfricke

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -32,6 +32,12 @@ try:
     from xgboost_ray.util import Event, Queue, MultiActorTask, \
         force_on_current_node
 
+    if ray.__version__ > "1.4.0":
+        # https://github.com/ray-project/ray/pull/16437
+        DEFAULT_PG = "default"
+    else:
+        DEFAULT_PG = None
+
     RAY_INSTALLED = True
 except ImportError:
     ray = get_node_ip_address = Queue = Event = ActorHandle = logger = None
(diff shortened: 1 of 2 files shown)


scv119:
alright.. looks it breaks a lot of tests...

rkooo567 (approved):
Maybe retry the build?

krfricke:
The change from `None` to `"default"` brings quite a bit confusion in here. I'm adjusting the PR to make it work again, but generally a non-backwards compatible API change in the placement group behavior is not great.

PR #123, opened 2021-06-18
Fix BATCH sharding

Yard1:
Fixes batch loading and petastorm loader not respecting indices

### DIFF
--- a/xgboost_ray/matrix.py
+++ b/xgboost_ray/matrix.py
@@ -885,8 +885,10 @@ def _get_sharding_indices(sharding: RayShardingMode, rank: int,
                           num_actors: int, n: int):
     """Return indices that belong to worker with rank `rank`"""
     if sharding == RayShardingMode.BATCH:
-        start_index = int(math.floor(rank / num_actors) * n)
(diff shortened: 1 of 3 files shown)


xgboost_ray/matrix.py
```diff
@@ -885,8 +885,10 @@ def _get_sharding_indices(sharding: RayShardingMode, rank: int,
                           num_actors: int, n: int):
     """Return indices that belong to worker with rank `rank`"""
     if sharding == RayShardingMode.BATCH:
-        start_index = int(math.floor(rank / num_actors) * n)
-        end_index = int(math.floor(rank + 1 / num_actors) * n)
+        start_index = int(rank * math.ceil(n / num_actors))
+        end_index = int((rank + 1) * math.ceil(n / num_actors))
+        if end_index > n:
+            end_index -= 1
```
krfricke, line 891: Does this work with range?

krfricke: Maybe just `min(n, end_index)`?

krfricke (changes requested):

krfricke (approved):
This is great. Thanks!

PR #126, opened 2021-06-21
Enable passing of RayDMatrix to sklearn API

Yard1:


krfricke (commented):
Generally looks good, but one of the test suites is failing

Yard1:
@krfricke The test failure on Ray master seems to be unrelated

krfricke (approved):
Awesome!

Issue #131, opened 2021-07-07
Batch sharing on release test with 16 workers throws an exception


Yard1:
Looks like the data is None for at least one worker. Will investigate

PR #132, opened 2021-07-07
Better algorithm for batch sharding indices

Yard1:
Fixes https://github.com/ray-project/xgboost_ray/issues/131

The previous algorithm had an edge case where it was possible for one actor to get no indices. The new algorithm is based on [`numpy.array_split`](https://github.com/numpy/numpy/blob/v1.21.0/numpy/lib/shape_base.py#L739-L792) and guarantees that each actor will get indices.


krfricke (approved):
Looks great, can we add a quick unit test for the edge case?

PR #136, opened 2021-07-16
fix placement group timeout behavior

matthewdeng:
When placement group setup times out, we are expected to raise a `TimeoutError` and fail immediately. When `ray.wait` times out, the first object returned will be an _empty list_, not `None`. This PR fixes the logic so that a timeout will properly route to a `TimeoutError`.

**Testing Done:** Added a new `test_timeout` test in `test_colocation.py` that tries to request an invalid resource. Without this change, the code will hang in `_train` as it waits for the actors to be prepared.

### DIFF
--- a/xgboost_ray/tests/test_colocation.py
+++ b/xgboost_ray/tests/test_colocation.py
@@ -176,12 +176,13 @@ class TestColocation(unittest.TestCase):
                 num_samples=1,
             )
 
+    @patch("xgboost_ray.main.PLACEMENT_GROUP_TIMEOUT_S", 5)
     def test_timeout(self):
         """Checks that an error occurs when placement group setup times out."""
         with self.ray_start_cluster() as cluster:
             ray.init(address=cluster.address)
 
-            with pytest.raises(TimeoutError):
+            with self.assertRaises(TimeoutError):
                 train(
                     self.params,
                     RayDMatrix(self.x, self.y),
--- a/xgboost_ray/tests/test_colocation.py
+++ b/xgboost_ray/tests/test_colocation.py
@@ -176,6 +176,21 @@ class TestColocation(unittest.TestCase):
                 num_samples=1,
             )
 
+    def test_timeout(self):
+        """Checks that an error occurs when placement group setup times out."""
+        with self.ray_start_cluster() as cluster:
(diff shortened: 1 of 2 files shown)


xgboost_ray/tests/test_colocation.py
```diff
@@ -176,6 +176,21 @@ def inner_func(config):
                 num_samples=1,
             )
 
+    def test_timeout(self):
+        """Checks that an error occurs when placement group setup times out."""
+        with self.ray_start_cluster() as cluster:
+            ray.init(address=cluster.address)
+
+            with pytest.raises(TimeoutError):
```
Yard1, line 184: Nit
```suggestion
            with self.assertRaises(TimeoutError):
```

Yard1:
> Note: Currently the test will timeout after the default value of PLACEMENT_GROUP_TIMEOUT_S which is 100 seconds. I wasn't sure how to properly override os.environ["PLACEMENT_GROUP_TIMEOUT_S"] to a smaller test value and propagate this into the evaluation of PLACEMENT_GROUP_TIMEOUT_S so that the test can terminate quicker.

@matthewdeng Try using mock.patch? Something like this:
```python
    @patch.dict(os.environ, {"PLACEMENT_GROUP_TIMEOUT_S": 5})
    def test_timeout(self):
```

matthewdeng:
> @matthewdeng Try using mock.patch? Something like this:
> 
> ```python
>     @patch.dict(os.environ, {"PLACEMENT_GROUP_TIMEOUT_S": 5})
>     def test_timeout(self):
> ```

Ah this won't work because `PLACEMENT_GROUP_TIMEOUT_S` is already resolved before it gets to this test. Is there a way to directly patch the `PLACEMENT_GROUP_TIMEOUT_S` variable?

Yard1:
Ah, of course. I guess you would have to patch the import of xgboost_ray itself. It shouldn't affect other tests here.

krfricke (commented):
Looks good, happy to merge once Antoni's comments are addressed

matthewdeng:
@Yard1 @krfricke thanks, updated the PR!

Yard1 (approved):

PR #145, opened 2021-08-09
Update Ray Datasets import

Yard1:
Ray Datasets path has been changed from `ray.experimental.data` to `ray.data` in the nightlies, breaking support here. This PR updates the imports in `xgboost_ray`.


richardliaw (approved):

krfricke (approved):
Seems to be compatible with the recent release, I'm just wondering if this breaks compatibility with older Ray versions (<1.5). But it should be fine since Ray datasets is an optional data source anyway.

PR #149, opened 2021-08-17
Fix resource detection + better fallback, complain about unset num_actors

### LINKED ISSUE #146: Remove default value for num_actors
Currently per default `num_actors=4`. We should get rid of this and require users to always specify the number of actors.

### LINKED ISSUE #147: Fix cluster resource detection in multi-node clusters for number_actors < number_nodes
When starting 4 distributed actors on a 8 node cluster with 16 CPUs each (144 CPUs), we autodetect 36 CPUs and schedule a placement group with 4 bundles a 36 CPUs.

Instead we should be smart about this and only assign 16 CPUs per worker (and maybe throw a warning that the cluster will be underutilized).

cc @mmui would you be interested in taking this?

krfricke:
Closes #146 
Closes #147 

cc @Yard1

### DIFF
--- a/xgboost_ray/examples/simple_predict.py
+++ b/xgboost_ray/examples/simple_predict.py
@@ -3,7 +3,7 @@ import os
 from sklearn import datasets
 
 import xgboost as xgb
-from xgboost_ray import RayDMatrix, predict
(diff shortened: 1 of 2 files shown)


Yard1 (approved):
:+1:

PR #150, opened 2021-08-24
Ensure that Dask Scheduler is set for Ray-on-Dask

ijrsvt:
Closes https://github.com/ray-project/ray/issues/16743


Yard1:
@ijrsvt looks like you need to add wrapt as a dependency

Yard1 (approved):
Thanks!

PR #151, opened 2021-08-30
Make xgboost an extra dependency

Yard1:
This has been done in order to allow for lightgbm-ray to be installed without xgboost. The new install command for xgboost-ray will be `pip install "xgboost-ray[default]"`. If xgboost is not installed, `train` and `predict` methods will throw an ImportError.


krfricke (approved):

richardliaw:
(oops, can we merge)?

Yard1:
Needs to pass CI now ;)

Issue #152, opened 2021-09-14
Support for __partitioned__ as data source


fschlimb:
An API/protocol for exchanging distributed, partitioned data is being defined. The goal is to establish a uniform way of describing such data while making it possible to avoid unnecessary data copies and even zero copy and so avoid specialized implementations for different data containers (like RayDataSet, MLDataSet, modin etc). The protocol basically defines the "meta" data about the distribution and partitioning (tiling) of the data. The current proposal is here: https://github.com/IntelPython/DPPY-Spec/blob/draft/partitioned/Partitioned.md, an issue/discussion is here: https://github.com/IntelPython/DPPY-Spec/issues/3

Any feedback/input to the discussion/spec is welcome.

There are several implementations/PRs for several projects, like HeAT, modin, MLDataSet and DAL.

It would be nice to have xgboost_ray support this, too. I will also open a PR.

PR #153, opened 2021-09-14
Supporting __partitioned__ interface

fschlimb:
Initial implementation for #152 .

Once MLDataSet and modin have support for __partitioned__ in mainline their specialized implementation could be removed without limiting any functionality.

### DIFF
--- a/xgboost_ray/data_sources/partitioned.py
+++ b/xgboost_ray/data_sources/partitioned.py
@@ -1,0 +1,98 @@
+from typing import Any, Optional, Sequence, Dict, Union, Tuple
+
+from collections import defaultdict
+import pandas as pd
+import numpy as np
+
+import ray
+from ray import ObjectRef
+from ray.actor import ActorHandle
+
+from xgboost_ray.data_sources._distributed import \
+    assign_partitions_to_actors, get_actor_rank_ips
+from xgboost_ray.data_sources.data_source import DataSource, RayFileType
+from xgboost_ray.data_sources.object_store import ObjectStore
+from xgboost_ray.data_sources.pandas import Pandas
+from xgboost_ray.data_sources.numpy import Numpy
+
+
+class Partitioned(DataSource):
+    """Read from distributed data structure implementing __partitioned__.
+
+    __partitioned__ provides meta data about how the data is partitioned and
+    distributed across several compute nodes, making supporting objects them
+    suitable for distributed loading.
+    """
+    supports_central_loading = True
+    supports_distributed_loading = True
+
+    @staticmethod
+    def is_data_type(data: Any,
+                     filetype: Optional[RayFileType] = None) -> bool:
+        return hasattr(data, "__partitioned__")
+
+    @staticmethod
+    def load_data(
+            data: Any,  # __partitioned__ dict
+            ignore: Optional[Sequence[str]] = None,
+            indices = None,  # FIXME Optional[Union[Sequence[int], Sequence[ObjectRef]]] = None,
+            **kwargs) -> pd.DataFrame:
+
+        assert isinstance(data, dict), "Expected __partitioned__ dict"
+        assert len(indices) == 0 or not isinstance(indices[0], int)
+        _get = data["get"]
+
+        if indices is None or len(indices) == 0:
+            tiling = data["partition_tiling"]
+            ndims = len(tiling)
+            # we need tuples to access partitions in the right order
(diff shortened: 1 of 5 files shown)


xgboost_ray/matrix.py
```diff
@@ -414,6 +415,7 @@ def get_data_source(self) -> Type[DataSource]:
             try:
                 if source.is_data_type(self.data, self.filetype):
                     data_source = source
+                    print("Using data source", data_source)
```
Yard1, line 418: Remove?

krfricke, line 418: Remove debug output?

xgboost_ray/data_sources/partitioned.py
```diff
@@ -0,0 +1,98 @@
...
+
+        if isinstance(local_df[0], pd.DataFrame):
+            return Pandas.load_data(pd.concat(local_df, copy=False), ignore=ignore)
+        else:
+            return Numpy.load_data(np.concatenate(local_df), ignore=ignore)
+
+    # @staticmethod
+    # def convert_to_series(data: Any) -> pd.Series:
+    #     # Presumably data is 'raw' data, e.g. something DataSource can work on
+    #     return DataSource.convert_to_series(data)
```
krfricke, line 65: Remove?

xgboost_ray/data_sources/partitioned.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @staticmethod
+    def load_data(
+            data: Any,  # __partitioned__ dict
+            ignore: Optional[Sequence[str]] = None,
+            indices = None,  # FIXME Optional[Union[Sequence[int], Sequence[ObjectRef]]] = None,
+            **kwargs) -> pd.DataFrame:
+
+        assert isinstance(data, dict), "Expected __partitioned__ dict"
+        assert len(indices) == 0 or not isinstance(indices[0], int)
```
Yard1, line 42: I think a more informative message (like a line above) would be good here

krfricke (commented):
Thank you, this is a very thorough PR.

I just triggered CI so we can see if the tests pass. Can you also add the example to `run_ci_examples.sh` in the root directory?

xgboost_ray/data_sources/partitioned.py
```diff
@@ -0,0 +1,98 @@
...
+from xgboost_ray.data_sources.pandas import Pandas
+from xgboost_ray.data_sources.numpy import Numpy
+
+
+class Partitioned(DataSource):
+    """Read from distributed data structure implementing __partitioned__.
+
+    __partitioned__ provides meta data about how the data is partitioned and
+    distributed across several compute nodes, making supporting objects them
+    suitable for distributed loading.
```
Yard1, line 24: Could we link to the spec here as in https://github.com/ray-project/xgboost_ray/issues/152?

fschlimb: Yes, we could. I did not because we hope to find a better home for the spec (like in the pydata API universe or at least the main branch there) at some point. Let me add it here nevertheless, I will record implementations that we can scan for reference should we move the spec.

xgboost_ray/data_sources/partitioned.py
fschlimb: Hm, Besides being incorrect (e.g. if indices is None) I guess this was mostly a debug help to understand what's happening. I am not sure I know what the type annotations (as copied from other data sources) allow list entries to be integers and if this can happen at all here. Any thoughts?

Yard1: If this is just a debug thing, then I think we can just remove it

fschlimb: I think I understand this now, I will fix the type annotation and remove the assertion.

xgboost_ray/data_sources/partitioned.py
fschlimb: Done.

fschlimb:
> Thank you, this is a very thorough PR.
> 
> I just triggered CI so we can see if the tests pass. Can you also add the example to `run_ci_examples.sh` in the root directory?

CI doesn't seem start anymore.

krfricke:
Ah sorry, it seems I have to manually trigger it every time. Just did!

fschlimb:
I have no idea how my changes could make any other test fail.

Yard1:
Looks like tests are being flaky on their own

PR #158, opened 2021-09-21
Add API annotations

krfricke:


xgboost_ray/main.py
```diff
@@ -42,6 +43,7 @@
     RAY_INSTALLED = True
 except ImportError:
     ray = get_node_ip_address = Queue = Event = ActorHandle = logger = None
+    PublicAPI = DeveloperAPI = lambda fn: fn
```
Yard1, line 46: This will throw an exception with the keyword argument, as in `PublicAPI(stability="beta")`:
```
>>> f = lambda fn: fn
>>> @f(stability="beta")
... def a():
...     return
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: <lambda>() got an unexpected keyword argument 'stability'
```

xgboost_ray/sklearn.py
```diff
@@ -557,6 +561,8 @@ def get_num_boosting_rounds(self):
 
 RayXGBRFRegressor.__doc__ = _treat_estimator_doc(XGBRFRegressor.__doc__)
 
+PublicAPI(stability="beta")
```
Yard1, line 564: Should this be here?

Yard1 (changes requested):

xgboost_ray/main.py
Yard1: You can copy the dummy decorator in the sklearn.py

xgboost_ray/sklearn.py
krfricke: Ah that happened a few times, will fix in a sec...

xgboost_ray/main.py
krfricke: Thanks!

Yard1 (approved):
Thanks!

PR #170, opened 2021-11-05
Preserve ordering of columns on conversion to RayDMatrix

amogkam:
Conversion of a pandas dataframe to a RayDMatrix did not conserve the ordering of columns because this line would actually do a set difference: https://github.com/ray-project/xgboost_ray/blob/88c3188741650ac7c6e2383315b637c4462f65e7/xgboost_ray/matrix.py#L253

Instead we should conserve the order of the columns in the original dataframe.

Issue: https://discuss.ray.io/t/raydmatrix-reordering-dataframe-columns/3981

### DIFF
--- a/xgboost_ray/tests/test_matrix.py
+++ b/xgboost_ray/tests/test_matrix.py
@@ -43,6 +43,16 @@ class XGBoostRayDMatrixTest(unittest.TestCase):
         data = RayDMatrix(self.x, self.y)
         self.assertTrue(ray.get(same.remote(data, data)))
 
+    def testColumnOrdering(self):
+        """when excluding cols, the remaining col order should be preserved."""
+
+        cols = [str(i) for i in range(50)]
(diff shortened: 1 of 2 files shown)


richardliaw:
Is there a test we can write for this?

Yard1 (approved):
Happy to merge it with a test

amogkam:
done & done! PTAL!

xgboost_ray/tests/test_matrix.py
```diff
@@ -43,6 +43,16 @@ def same(one, two):
         data = RayDMatrix(self.x, self.y)
         self.assertTrue(ray.get(same.remote(data, data)))
 
+    def testColumnOrdering(self):
+        """when excluding cols, the remaining col order should be preserved."""
```
amogkam, line 47:
```suggestion
        """When excluding cols, the remaining col order should be preserved."""
```

Yard1 (approved):

krfricke (approved):

PR #171, opened 2021-11-05
Unpin pyarrow as petastorm was updated

Yard1:
Unpin pyarrow as petastorm was updated.


Yard1:
Test failures are unrelated and same on master

krfricke (approved):

PR #173, opened 2021-11-10
➕  Re-add pyarrow dependency to `install_requires`

TeoZosa:
# TL;DR

Fixes `ModuleNotFoundError: No module named 'pyarrow'` errors due to pyarrow  (accidentally?) being removed in #171.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,6 @@ setup(
     "distributed computing framework Ray.",
     url="https://github.com/ray-project/xgboost_ray",
     install_requires=[
-        "ray>=1.6", "numpy>=1.16,<1.20", "pandas", "wrapt>=1.12.1",
+        "ray>=1.6", "numpy>=1.16,<1.20", "pandas", "pyarrow", "wrapt>=1.12.1",
         "xgboost>=0.90"
     ])
--- a/setup.py
+++ b/setup.py
@@ -13,4 +13,3 @@ setup(
         "ray>=1.6", "numpy>=1.16,<1.20", "pandas", "wrapt>=1.12.1",
         "xgboost>=0.90"
     ])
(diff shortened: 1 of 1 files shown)


TeoZosa:
> When does this error come up? We shouldn't depend on pyarrow per se, just some data sources might require it.

Hey @krfricke, thanks for following up on this. It looks like that's the case. Truncating the traceback in fb16ed4, it looks like importing `xgboost_ray` automatically imports a data source which in turn imports a pyarrow module down the line:

```
Traceback (most recent call last):
...
    import xgboost_ray as xgbr
...
    from xgboost_ray.main import RayParams, train, predict
...
    from xgboost_ray.matrix import RayDMatrix, combine_data, \
...
    from xgboost_ray.data_sources import DataSource, data_sources, RayFileType
...
    from xgboost_ray.data_sources.ml_dataset import MLDataset
...
    from ray.util.data import MLDataset as MLDatasetType
...
    from ray.util.data.parquet import read_parquet
...
    import pyarrow.parquet as pq
ModuleNotFoundError: No module named 'pyarrow'
```

Yard1:
Yeah, sorry, I didn't notice your PR when I made mine @TeoZosa. @krfricke Should we make it a soft import instead?

krfricke:
Thanks for the context, yes, we should make this a soft import, and if the requirements are not met we should raise an actionable error message (i.e. mentioning the `pip install` command required to fix this).

TeoZosa:
> Yeah, sorry, I didn't notice your PR when I made mine @TeoZosa.

All good! I figure you guys would have more background to coordinate the changes in the best way (which it sounds like is indeed the case :) )

Yard1:
@krfricke I can update my PR with that, then

PR #174, opened 2021-11-10
📌  Unpin numpy max version (remove ~<1.20~)

TeoZosa:
# ⚠️ TODO: blocked on merge of #173

# TL;DR

Unpins numpy's max version (to, in my case, fix compatibility with M1 macs).


pablete:
Currently ray[data]  requires numpy >= 1.20
while xgboost-ray==0.1.5  requires numpy < 1.20

+1 for this change

krfricke (approved):

PR #180, opened 2021-12-11
Fix broken examples

Yard1:
We aren't testing higgs ones in CI, so this slipped through the cracks.


krfricke (approved):
Thanks! Should we add this to the CI then?

PR #185, opened 2022-01-20
Fix cutting edge CI

Yard1:


krfricke (approved):
Thanks! Seems like a performance degradation in xgboost 1.6.0?

Yard1:
Looks like it, yeah

PR #193, opened 2022-02-16
Fix "qid" being passed to legacy DMatrix

Yard1:


amogkam (approved):
Thanks for the quick follow up here!

Can we make sure to test this code path in the legacy test suite in a follow up PR?

PR #203, opened 2022-03-17
Mitigate finishing too quickly

Yard1:
Same as in https://github.com/ray-project/lightgbm_ray/pull/29

This doesn't seem to occur with XGBoost, but the same logic is introduced for consistency. `test_tune` has been similarly deflaked.


krfricke (approved):
LGTM

PR #204, opened 2022-03-25
Adjust readme for Ray Docs

Yard1:
Adjust the readme file to take advantage of https://github.com/ray-project/ray/pull/23505

There are no user facing changes, the contents of the readme displayed on GitHub are the same as they were before.


krfricke (approved):
Let's try it without the table conversion, but otherwise great!

Issue #208, opened 2022-04-05
ValueError: cannot find context for 'fork' on windows


tfgstudios:
Running this code doesn't work
`
            model = xgboost_ray.train(
                {
                    # This is important to be set correctly for the problem type (multi class in our case)
                    'objective': objective,

                    'eval_metric': self.val_metric,
                    'enable_categorical': enable_categorical,
                    'use_label_encoder': use_label_encoder,
                    'tree_method': tree_method,
                    'max_depth': max_depth,
                    'n_jobs': n_jobs,
                    'learning_rate': learning_rate,
                    'num_parallel_tree': num_parallel_tree,
                    'colsample_bylevel': colsample_bylevel,

                    'sample_weight': weights.train.samples,
                    'sample_weight_eval_set': [weights.val.samples],
                    'feature_weights': weights.flatten_sequence_weights,

                    'min_child_weight': min_child_weight,
                    'gamma': gamma,
                    'reg_alpha': reg_alpha,
                    'reg_lambda': reg_lambda,
                    'max_bin': max_bin,

                    'random_state': random_state,
                    'verbosity': verbosity,
                },
                train_set,
                num_boost_round=num_boost_round,
                evals=[(val_set, 'val')],
                early_stopping_rounds=self.early_stopping_rounds,
                verbose_eval=10,

                # Ray-specific params
                ray_params=RayParams(
                    num_actors=1,
                    cpus_per_actor=max(1, os.cpu_count()-1),
                    gpus_per_actor=-1,
                )
            )`

Tried to set `multiprocessing.set_start_method("spawn", force=True)` before, but it doesn't work

amogkam:
Hey @tfgstudios, thanks for raising this issue! Unfortunately `xgboost-ray` does not work on Windows. We require the `fork` method to be used for multiprocessing instead of `spawn`. 

We will add this as an enhancement to provide support for this in the future!

amogkam:
Going to close this issue, but @tfgstudios please +1 this issue for adding windows support #210!

PR #42, opened 2021-01-06
Add 3.8 Tests and force fork strategy

amogkam:

PR #209, opened 2022-04-06
Fail quickly on Windows

amogkam:
`xgboost-ray` training currently does not work on Windows. This is because we use multiprocessing in the Rabit Tracker, but the run method is not serializable, so we have to use `fork` method instead of `spawn`. (see #42). 

However, Python multiprocessing does not support `fork` method on Windows.

This PR immediately raises an error if Windows OS is identified with a better error message.

See #208

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -1,4 +1,4 @@
-import sys
+import platform
 from typing import Tuple, Dict, Any, List, Optional, Callable, Union, Sequence
 from dataclasses import dataclass, field
 from distutils.version import LooseVersion
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -1,3 +1,4 @@
+import sys
 from typing import Tuple, Dict, Any, List, Optional, Callable, Union, Sequence
 from dataclasses import dataclass, field
 from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


xgboost_ray/main.py
```diff
@@ -198,7 +199,8 @@ def start(self, nworker):
         multiprocessing.set_start_method("fork", force=True)
 
         def run():
-            self.accept_workers(nworker)
+            self
```
Yard1, line 202: Revert changes?

amogkam: Oops yeah, reverted!

xgboost_ray/main.py
```diff
@@ -1,3 +1,4 @@
+import sys
```
Yard1, line 1:
```suggestion
import platform
```

xgboost_ray/main.py
```diff
@@ -1230,6 +1231,10 @@ def train(
     """
     os.environ.setdefault("RAY_IGNORE_UNHANDLED_ERRORS", "1")
 
+    if sys.platform.startswith("win"):
```
Yard1, line 1234:
```suggestion
    if platform.system() == "Windows":
```

Yard1 (commented):
Actually can we use `platform` instead? It should be the most robust

Yard1 (approved):

amogkam:
@Yard1 @krfricke is CI supposed to be passing? Not sure what's going on here since the change should be unrelated.

Yard1:
@amogkam It's a bit flaky atm. The CI looks good, let me merge.

PR #211, opened 2022-04-08
Fix `test_num_parallel_tree` for cutting edge CI

Yard1:
Also uses `LooseVersion` instead of our own regex-based version tuple.


krfricke (approved):
Thanks!

PR #212, opened 2022-04-19
Fix CI failing with `xgboost==1.6.0`

Yard1:


xgboost_ray/tests/test_sklearn.py
```diff
@@ -313,7 +313,7 @@ def test_num_parallel_tree(self):
             dump = bst.get_booster().get_dump(dump_format="json")
             assert len(dump) == 4
 
-            if XGBOOST_LOOSE_VERSION >= LooseVersion("2.0.0"):
+            if XGBOOST_LOOSE_VERSION >= LooseVersion("1.6.0"):
                 config = json.loads(bst.get_booster().save_config())
                 assert (int(config["learner"]["gradient_booster"][
                     "gbtree_model_param"]["num_parallel_tree"]) == 4)
```
trivialfis, line 319: Out of curiosity, would it be useful if we expose this as a property?

Yard1: `XGBOOST_LOOSE_VERSION` or `"1.6.0"`?

krfricke: I think @trivialfis means the number of trees. Yes, that would actually be very helpful, we have a few tests that use this

krfricke (approved):

xgboost_ray/tests/test_sklearn.py
Yard1: Ah, gotcha.

PR #217, opened 2022-05-12
Remove code for legacy Ray versions

krfricke:
There are several places where we still have code paths for legacy Ray versions. With the next release we can require newer Ray versions to be used.


Yard1 (approved):
Thanks, this looks great! I think LightGBM-Ray will also need to be updated. I will make a PR after this is merged.

PR #218, opened 2022-05-20
Remove support for deprecated `ray.util.data.MLDataset`

amogkam:
Ray MLDataset has been deprecated. We should no longer maintain support for it in newer xgboost-ray versions.


amogkam:
cc @jjyao

jjyao (approved):

Yard1 (approved):

PR #226, opened 2022-07-23
Reduce verbosity for ray events

richardliaw:
Adds a verbose flag to RayParams to toggle events. Defaults to False, but open to discuss.

The main thing is that when using AIR/Tune, this logging becomes quite verbose.

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -384,7 +384,7 @@ class RayParams:
     # Distributed callbacks
     distributed_callbacks: Optional[List[DistributedCallback]] = None
 
-    verbose: bool = False
+    verbose: Optional[bool] = None
 
     def get_tune_resources(self):
         """Return the resources to use for xgboost_ray training with Tune."""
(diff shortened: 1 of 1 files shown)


xgboost_ray/main.py
```diff
@@ -382,6 +384,8 @@ class RayParams:
     # Distributed callbacks
     distributed_callbacks: Optional[List[DistributedCallback]] = None
 
+    verbose: bool = False
```
krfricke, line 387:
```suggestion
    verbose: bool = True
```

can we default this to True for non-tune runs and to False for tune runs (which includes Ray AIR)?

krfricke (approved):
Generally lgtm but one question for default

xgboost_ray/main.py
richardliaw: Hmm, how do you do that?

Yard1:
CI should be fixed by https://github.com/ray-project/xgboost_ray/pull/228

PR #232, opened 2022-09-01
Switch to use packaging.Version from distutils LooseVersion

peytondmurray:
## Summary

This PR removes uses of `distutils.version.LooseVersion` in favor of `packaging.version.Version` at the suggestion of the deprecation warning generated when a user makes use of `LooseVersion`. This PR addresses one of the deprecation warnings identified in https://github.com/ray-project/ray/issues/27851. `distutils` [will be deprecated in Python 3.10 and 3.11, and removed entirely in 3.12](https://peps.python.org/pep-0632/).

## Changes

- Replaced `distutils.version.Version` with `packaging.version.Version` everywhere
- Added `packaging` as a dependency in `requirements-test.txt` and in `setup.py`

## Tests

I tested this on my local machine with `pytest` and all the tests that weren't skipped passed.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -2,6 +2,7 @@ flake8==3.7.7
 flake8-comprehensions
 flake8-quotes==2.0.0
 yapf==0.23.0
+packaging
 petastorm
 pytest
 pyarrow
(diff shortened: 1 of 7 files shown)


Yard1 (approved):
Thanks! Would you be able to do the same for lightgbm-ray?

peytondmurray:
> Thanks! Would you be able to do the same for lightgbm-ray?

Definitely, I'll take a look later today!

Looks like there's a formatting issue - I'll run the formatter and push again.

Yard1:
Thanks!

PR #234, opened 2022-09-15
add license metadata tag to package setup

jimthompson5802:
Add license metadata tag to package setup 

package scanning system prevents install of packages with missing license metadata.  Unable to test this package.


Yard1 (approved):
Thanks!

PR #235, opened 2022-09-19
Add `enable_categorical`, better detection for being in a PG

Yard1:
XGBoost>=1.5 added a new dmatrix param, `enable_categorical`. Adding it here.

This PR also makes the detection of the case of already being in a placement group more broad and not limited to just Tune. This allows for xgboost-ray to be ran in nested tasks that are not related to Tune (this should be two PRs really, but I needed one branch to patch it for hackathon).


krfricke (approved):
Thanks!

krfricke:
Hm, maybe we should add a test for the enable_categorical property?

Yard1:
@krfricke we can test that in AIR with a predictor?

PR #239, opened 2022-10-03
Add sort dataframe logic on qid

atomic:
If qid is provided, xgb.DMatrix [requires data to be sorted by qid - see data.cc](https://github.com/dmlc/xgboost/blob/master/src/data/data.cc#L486):
https://github.com/ray-project/xgboost_ray/blob/master/xgboost_ray/matrix.py#L351

draft to add auto sorting of dataframe if qid is given and dataframe is not already sorted by qid

### DIFF
--- a/xgboost_ray/matrix.py
+++ b/xgboost_ray/matrix.py
@@ -48,6 +48,25 @@ def concat_dataframes(dfs: List[Optional[pd.DataFrame]]):
     return pd.concat(filtered, ignore_index=True, copy=False)
 
 
+def ensure_sorted_by_qid(df: pd.DataFrame, qid: Data) -> pd.DataFrame:
+    _qid = None
+    if isinstance(qid, str):
+        _qid = df[qid]
+    elif isinstance(qid, np.ndarray):
+        _qid = pd.Series(qid)
+    elif isinstance(qid, pd.DataFrame):
+        _qid = qid.iloc[:, 0]
+    elif isinstance(qid, pd.Series):
+        _qid = qid
+    if _qid.is_monotonic:
+        return df
+    else:
+        if isinstance(qid, str):
+            return df.sort_values([qid])
+        else:  # case when qid is not part of df
+            return df.set_index(_qid).sort_index().reset_index(drop=True)
+
+
 @PublicAPI(stability="beta")
 class RayShardingMode(Enum):
     """Enum for different modes of sharding the data.
@@ -228,8 +247,8 @@ class _RayDMatrixLoader:
 
         """
         # sort dataframe by qid if exists (required by DMatrix)
-        if self.qid is not None and not local_data[self.qid].is_monotonic:
-            local_data = local_data.sort_values([self.qid])
+        if self.qid is not None:
+            local_data = ensure_sorted_by_qid(local_data, self.qid)
 
         exclude_cols: Set[str] = set()  # Exclude these columns from `x`
 
--- a/xgboost_ray/matrix.py
+++ b/xgboost_ray/matrix.py
@@ -227,6 +227,10 @@ class _RayDMatrixLoader:
         `label_upper_bound`
 
         """
+        # sort dataframe by qid if exists (required by DMatrix)
+        if self.qid and not local_data[self.qid].is_monotonic:
+            local_data = local_data.sort_values([self.qid])
+
         exclude_cols: Set[str] = set()  # Exclude these columns from `x`
 
         label, exclude = data_source.get_column(local_data, self.label)
@@ -349,11 +353,6 @@ class _CentralRayDMatrixLoader(_RayDMatrixLoader):
         # yet. Instead, we'll be selecting the rows below.
         local_df = data_source.load_data(
             self.data, ignore=self.ignore, indices=None, **self.kwargs)
-
-        # sort dataframe by qid if exists (required by DMatrix)
-        if self.qid and not local_df[self.qid].is_monotonic:
-            local_df = local_df.sort_values([self.qid])
-
         x, y, w, b, ll, lu, qid = self._split_dataframe(
             local_df, data_source=data_source)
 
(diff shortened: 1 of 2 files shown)


Yard1:
I wonder if it wouldn't work better to include this logic in `_get_dmatrix` function in `main.py`. X and y there are pandas objects, so we can sort X and then apply the index to y. That way we support all cases as the sorting would happen on each actor right before a DMatrix is created.

atomic:
@Yard1 I don't think we can sort cleanly in `_get_dmatrix` since the columns are already selected out of the dataframe to list.

Yard1:
`_split_dataframe` works for me! Can we encapsulate the logic in a separate method for better modularity and call it in `_split_dataframe`? Thanks, let's see if CI is happy :)

Can we also add a test that would fail without sorted qids? The existing one can be modified to add a case like that.

xgboost_ray/matrix.py
```diff
@@ -227,6 +227,10 @@ def _split_dataframe(
         `label_upper_bound`
 
         """
+        # sort dataframe by qid if exists (required by DMatrix)
+        if self.qid and not local_data[self.qid].is_monotonic:
```
Yard1, line 231:
```suggestion
        if self.qid is not None and not local_data[self.qid].is_monotonic:
```
should fix test failures!

xgboost_ray/matrix.py
```diff
@@ -227,6 +227,10 @@ def _split_dataframe(
...
+            local_data = local_data.sort_values([self.qid])
```
heyitsmui, line 232: this is awesome! so this should be done at each worker, after each parquet is loaded into pdarray and concatenated into `local_data` and then you sort it by `qid`? After that you will use it to create `dmatrix` and do a `ray.put` to object store?

heyitsmui: @atomic can you also add a few unit test cases that cover our use case e.g. qids are across multiple parquet files and as long as we make sure per-worker level sorting works this will work?

xgboost_ray/matrix.py
```diff
@@ -48,6 +48,25 @@ def concat_dataframes(dfs: List[Optional[pd.DataFrame]]):
...
 
 
+def ensure_sorted_by_qid(df: pd.DataFrame, qid: Data) -> pd.DataFrame:
+    _qid = None
+    if isinstance(qid, str):
+        _qid = df[qid]
+    elif isinstance(qid, np.ndarray):
+        _qid = pd.Series(qid)
+    elif isinstance(qid, pd.DataFrame):
+        _qid = qid.iloc[:, 0]
```
Yard1, line 58: can we raise an exception if it has more than 1 column?

atomic: sure

xgboost_ray/matrix.py
atomic: yes, this sorting is done after ray concat all the actor shard's loaded dataframe into one. And yes, the sorted data is then put to ray object store.

atomic: @heyitsmui added integ tests for multi-parquet files

xgboost_ray/tests/test_matrix.py
```diff
@@ -356,6 +356,56 @@ def testLegacyParams(self):
...
+            })
+        _ = DMatrix(**{
+            "data": in_x,
+            "label": in_y,
+            "qid": np.sort(unsorted_qid)
+        })  # no exception
+        # test RayDMatrix handles sorting automatically
+        mat = RayDMatrix(in_x, in_y, qid=unsorted_qid)
+        params = mat.get_data(rank=0, num_actors=1)
+        _ = DMatrix(**params)
```
Yard1, line 380:
```suggestion
            DMatrix(**{
                "data": in_x,
                "label": in_y,
                "qid": unsorted_qid
            })
        DMatrix(**{
            "data": in_x,
            "label": in_y,
            "qid": np.sort(unsorted_qid)
        })  # no exception
        # test RayDMatrix handles sorting automatically
        mat = RayDMatrix(in_x, in_y, qid=unsorted_qid)
        params = mat.get_data(rank=0, num_actors=1)
        DMatrix(**params)
```

xgboost_ray/tests/test_matrix.py
```diff
@@ -356,6 +356,56 @@ def testLegacyParams(self):
...
+            data_df["label"] = pd.Series(self.y)
+            data_df["group"] = pd.Series(unsorted_qid2)
+            data_df.to_parquet(parquet_file2)
+            mat = RayDMatrix([parquet_file1, parquet_file2],
+                             columns=["a", "b", "c", "d",
+                                      "label", "group"],
+                             label="label",
+                             qid="group")
+            params = mat.get_data(rank=0, num_actors=1)
+            _ = DMatrix(**params)
```
Yard1, line 407:
```suggestion
            DMatrix(**params)
```

Yard1 (approved):
Hey @atomic this is great! 2 small nits regarding tests. Can you also run `format.sh` script in the root repo folder to make lint CI pass? Thanks!

Yard1:
@atomic `xgboost_ray/matrix.py:61:80: E501 line too long (99 > 79 characters)` (just this left!)

xgboost_ray/tests/test_matrix.py
```diff
@@ -375,6 +380,8 @@ def testQidSortedBehaviorXGBoost(self):
         params = mat.get_data(rank=0, num_actors=1)
         DMatrix(**params)
 
+    @unittest.skipIf(XGBOOST_VERSION < Version("0.9.0"),
```
Yard1, line 383:
```suggestion
    @unittest.skipIf(XGBOOST_VERSION <= Version("0.9.0"),
```

xgboost_ray/tests/test_matrix.py
```diff
@@ -356,6 +359,8 @@ def testLegacyParams(self):
             label_lower_bound=label_lower_bound,
             label_upper_bound=label_upper_bound)
 
+    @unittest.skipIf(XGBOOST_VERSION < Version("0.9.0"),
```
Yard1, line 362:
```suggestion
    @unittest.skipIf(XGBOOST_VERSION <= Version("0.9.0"),
```

PR #259, opened 2023-01-13
Improve Ray Datasets support

Yard1:
Slightly improves performance and majorly improves object store memory usage with Ray Datasets. Also replaces our custom logic with Ray Datasets `split` method with locality hints - meaning the behavior should be the same as in Ray Train.


krfricke (approved):
Awesome!

PR #265, opened 2023-02-04
Add feature weights support

mozjay0619:
This PR adds support for feature_weights parameter, which is xgb.DMatrix parameter since Xgboost version 1.3.0. This PR does not add this support for RayDeviceQuantileDMatrix class.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import find_packages, setup
 setup(
     name="xgboost_ray",
     packages=find_packages(where=".", include="xgboost_ray*"),
-    version="0.1.14",
+    version="0.1.15",
     author="Ray Team",
     description="A Ray backend for distributed XGBoost",
     license="Apache 2.0",
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from setuptools import find_packages, setup
 setup(
     name="xgboost_ray",
     packages=find_packages(where=".", include="xgboost_ray*"),
-    version="0.1.15",
+    version="0.1.14",
     author="Ray Team",
     description="A Ray backend for distributed XGBoost",
     license="Apache 2.0",
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ project-id
 *.csv
 *.csv.gz
 *.parquet
+
+# Byte-compiled files
+__pycache__/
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 setup(
     name="xgboost_ray",
     packages=find_packages(where=".", include="xgboost_ray*"),
-    version="0.1.14",
+    version="0.1.15",
```
Yard1, line 6: There is no need to change the version in this PR.

Yard1 (commented):
Thanks for the PR! Can you run `format.sh` in the root folder to make sure the code is properly formatted?

setup.py
mozjay0619: Yeah, thanks for the comment. Will do!

mozjay0619:
@Yard1 I'm not sure how to proceed with the latest pytest failure. May be this is a temporary issue on Azure repo side? The error message says ``Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?``, but I am not sure how to do this. Could you please advise? Thanks!

Yard1:
@mozjay0619 I just reran the job and it passed that point. It looks like a random failure not related to your changes.

Yard1:
hmm @mozjay0619 we seem to have some actual errors, can you check?

mozjay0619:
@Yard1 Sorry, let me explain the nature of the error since we've had this multiple times. The feature_weights leverages the colsample feature, so it is inherently stochastic. The weights are configured to be monotonically increasing with the last (9th) feature having the biggest contribution. But due to the stochasticity, the second to last feature sometimes overpowers the last feature. I increased the boosting rounds for convergence and lowered the colsample rates as well. And this time, I tested it against 100 random seed values. I can take more drastic measures but I don't want to alter the code too much from the original demo code provided by xgboost official doc...

Yard1:
@mozjay0619 can we also set the seed as a parameter for xgboost? The numpy seed will not be carried over to the xgboost-ray Actors carrying out the training.

mozjay0619:
@Yard1 Yeah, I had entertained that exact idea. But I couldn't find an easy way to do it since the xgboost.train() method does not have seed parameter, and it relies on numpy random state. I suspect this is why the demo code sets numpy random state. One hacky way might be to create a seed parameter to ``train`` method in xgboost_ray and use that number to set the local seed for each Actor? But I'm not sure what this would actually do in the context of distributed xgboost training...

Yard1:
I believe it has a `seed` learning parameter but if the test passes here then I am fine with that. Thanks for iterating on this!

Yard1 (approved):

PR #267, opened 2023-02-13
Fix modin import with mismatched pandas version

Yard1:
Signed-off-by: Antoni Baum <antoni.baum@protonmail.com>


krfricke (approved):
Thanks for the quick fix!

PR #274, opened 2023-03-03
Fix _ray_get_act_cpus and set nthread in param passed to DMatrix creation

### LINKED ISSUE #266: Long single-core period during training hurts performance
My script is here:
```python
xgb_params = {
    'tree_method': 'hist',
    'grow_policy': 'depthwise'
}
cpus_per_actor=32
bst = xgboost_ray.train(
    params=xgb_params,
    dtrain=dtrain,
    num_boost_round=100,
    ray_params=xgboost_ray.RayParams(
        max_actor_restarts=0,
        gpus_per_actor=0,
        cpus_per_actor=cpus_per_actor,
        num_actors=16
    )
 )
```
I checked the CPU utilization using top during training, and I notice that there is a long time when CPU utilization is 100% after the training actors start, about 300s. Total training time is over 500s. Then the utilization goes up to around 2000%. I used py-spy to dump the stack trace during the long single-core period, and found this function `xgboost::tree::QuantileHistMaker::Builder::InitData` took a long time to finish. 

I am not sure what this function is doing, and why it is single core, but it is now the bottleneck of my workload. Any help would be appreciated, Thanks

kira-lin:
This pr closes #266

_ray_get_act_cpus  did not get the correct number set for CPU in actors, because when placement group is used, the resource name changes. In addition, set this number as nthread in param which is used to create DMatrix. @Yard1

### DIFF
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -296,7 +296,7 @@ def _ray_get_actor_cpus():
         for key in resource_ids.keys():
             if key.startswith("CPU"):
                 return resource_ids[key]
-        return None
+        return 1
 
 
 def _ray_get_cluster_cpus():
--- a/xgboost_ray/main.py
+++ b/xgboost_ray/main.py
@@ -293,7 +293,10 @@ def _ray_get_actor_cpus():
             return sum(cpu[1] for cpu in resource_ids["CPU"])
     else:
         resource_ids = ray.get_runtime_context().get_assigned_resources()
-        return resource_ids.get("CPU")
+        for key in resource_ids.keys():
+            if key.startswith("CPU"):
+                return resource_ids[key]
+        return None
 
 
 def _ray_get_cluster_cpus():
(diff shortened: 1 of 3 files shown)


xgboost_ray/main.py
```diff
@@ -599,6 +602,8 @@ def load_data(self, data: RayDMatrix):
         else:
             self._local_n[data] = len(param["data"])
 
+        # set nthread for dmatrix conversion
+        param["nthread"] = int(_ray_get_actor_cpus())
```
Yard1, line 606: this is going to fail if `_ray_get_actor_cpus` returns `None` (as that cannot be converted to int)

Yard1 (commented):
Thanks, this is great! Could we get a quick unit test for this? just checking if `_ray_get_actor_cpus()` returns the expected value.

Yard1 (approved):

Yard1:
Thanks!

PR #281, opened 2023-04-24
Reformat using black

krfricke:
The Ray main repo is using black, so we may want to do the same.


Yard1 (approved):

Yard1:
Please give lightgbm-ray the same treatment :)

PR #287, opened 2023-07-24
Update Ray Dataset with prediction example

krfricke:
Prediction with Ray Data doesn't work with distributed loading (see https://discuss.ray.io/t/raytaskerror-typeerror/11486/2). This PR adds a simple example on how to do batch inference with Ray Data.

Ideally we can convert this automatically to Ray Data-based batch inference in `predict()` in the short term. In the long term, we should discontinue all non-ray-data APIs and move data source support to ray data instead (except for petastorm, most of them should be supported anyways).

### DIFF
--- a/xgboost_ray/examples/simple_ray_dataset.py
+++ b/xgboost_ray/examples/simple_ray_dataset.py
@@ -24,7 +24,7 @@ def main(cpus_per_actor, num_actors):
     data.columns = [str(c) for c in data.columns]
     data["label"] = y
 
-    ray_ds = ray.data.from_pandas(data).repartition(num_actors)
+    ray_ds = ray.data.from_pandas(data)
     train_set = RayDMatrix(ray_ds, "label")
 
     evals_result = {}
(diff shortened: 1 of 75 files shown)


xgboost_ray/examples/simple_ray_dataset.py
```diff
@@ -22,16 +24,7 @@ def main(cpus_per_actor, num_actors):
...
-    # and current Ray master API
-    if hasattr(ray.data, "from_pandas_refs"):
-        # Generate Ray dataset from 4 partitions
-        ray_ds = ray.data.from_pandas(data).repartition(num_actors)
-    else:
-        # Split into 4 partitions
-        partitions = [ray.put(part) for part in np.split(data, num_actors)]
-        ray_ds = ray.data.from_pandas(partitions)
-
+    ray_ds = ray.data.from_pandas(data).repartition(num_actors)
```
Yard1, line 27: do we even need to do repartition anymore? this should be handled inside xgboost-ray nowadays, no?

Yard1 (approved):
thanks, one question!

xgboost_ray/examples/simple_ray_dataset.py
krfricke: You're right - removed

PR #296, opened 2023-09-13
Compatibility with XGBoost 2.0

krfricke:


xgboost_ray/sklearn.py
```diff
@@ -863,8 +849,9 @@ def _ray_fit_preprocess(self, y) -> Callable:
             ):
                 raise ValueError(label_encoding_check_error)
         else:
-            self.classes_ = np.unique(y)
-            self.n_classes_ = len(self.classes_)
+            if XGBOOST_VERSION < Version("2.0.0"):
+                self.classes_ = np.unique(y)
+            self.n_classes_ = len(np.unique(y))
```
trivialfis, line 854:
```suggestion
            self.n_classes_ = len(self.classes_)
```

trivialfis (commented):
Feel free to ping me if there's anything I can help.

xgboost_ray/sklearn.py
krfricke: Isn't this recursive? As

```
    @property
    def classes_(self) -> np.ndarray:
        return np.arange(self.n_classes_)
```

Yard1 (approved):
lgtm

